PR #1, opened 2019-06-10
Xen driver init

Wenzel:
Add Xen driver with basic init and pause/resume implementations.

cc @tathanhdinh, if you would like to review the code

Dependencies:
https://github.com/Wenzel/xenctrl
https://github.com/Wenzel/xen-sys (should be renamed to `xenctrl-sys`)
https://github.com/Wenzel/xenstore
https://github.com/Wenzel/xenstore-sys

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,3 +6,4 @@ edition = "2018"
 
 [dependencies]
 xenctrl = { version = "0.1.0", path = "../xenctrl" }
+xenstore = { version = "0.1.0", path = "../xenstore" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,4 +5,7 @@ authors = ["Mathieu Tarral <mathieu.tarral@protonmail.com>"]
 edition = "2018"
 
 [dependencies]
-xenctrl-sys = "0.1.0"
+xenctrl-sys = { version = "0.1.0", optional = true }
+
+[features]
+xen = ["xenctrl-sys"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,3 +5,4 @@ authors = ["Mathieu Tarral <mathieu.tarral@protonmail.com>"]
 edition = "2018"
 
 [dependencies]
+xen-sys = { path = "../xen-sys" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -5,7 +5,4 @@ authors = ["Mathieu Tarral <mathieu.tarral@protonmail.com>"]
 edition = "2018"
 
 [dependencies]
-xenctrl-sys = { version = "0.1.0", optional = true }
-
-[features]
-xen = ["xenctrl-sys"]
+xenctrl = { version = "0.1.0", path = "../xenctrl" }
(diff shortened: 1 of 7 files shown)


Wenzel:
I could use the right values in these Enum declarations:
https://github.com/Wenzel/xenstore/blob/master/src/lib.rs#L9

I got a compile error:
~~~rust
pub enum XBTransaction {
    Null = xenstore_sys::XBT_NULL,
}
~~~
~~~
[vagrant@localhost libmicrovmi]$ cargo build
   Compiling xenstore v0.1.0 (/vagrant/rust/xenstore)
error[E0308]: mismatched types
 --> /vagrant/rust/xenstore/src/lib.rs:9:12
  |
9 |     Null = xenstore_sys::XBT_NULL,
  |            ^^^^^^^^^^^^^^^^^^^^^^ expected isize, found u32

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: Could not compile `xenstore`.
~~~

Any ideas ?

Also here, how to match a null pointer ?
~~~rust
    pub fn new(open_type: XsOpenFlags) -> Result<Self,Error> {
        let xs_handle = unsafe {
            xenstore_sys::xs_open(open_type as u64)
        };
        match xs_handle {
            null_mut() => Err(Error::last_os_error()),
            _ => Ok(Xs {
                handle: xs_handle,
            }),
        }
    }
~~~
~~~
[vagrant@localhost libmicrovmi]$ cargo build
   Compiling xenstore v0.1.0 (/vagrant/rust/xenstore)
error[E0532]: expected tuple struct/variant, found function `null_mut`
  --> /vagrant/rust/xenstore/src/lib.rs:31:13
   |
31 |             null_mut() => Err(Error::last_os_error()),
   |             ^^^^^^^^ not a tuple struct/variant

error: aborting due to previous error

For more information about this error, try `rustc --explain E0532`.
error: Could not compile `xenstore`.
~~~

tathanhdinh:
@Wenzel 
The first error is because of the underlying type of enum is `isize`, as [described](https://doc.rust-lang.org/reference/items/enumerations.html):

> Under the default representation, the specified discriminant is interpreted as an isize value...

It can be fixed by, either type conversion:

```rust
pub enum XBTransaction {
    Null = xenstore_sys::XBT_NULL as _,
}
```

or change the underlying type

```rust
#[repr(u32)]
pub enum XBTransaction {
    Null = xenstore_sys::XBT_NULL,
}
```

The second error is because a [call expression](https://doc.rust-lang.org/reference/expressions/call-expr.html) cannot be used in a match arm (as a pattern). It can be fixed by, either

- depending on `xs_handle`'s type (`*const _` or `*mut _`): checking directly `xs_handle.is_null()`, or
- if...then expression

```rust
if xs_handle == null_mut() {
  ...
```

- if you still want to use matching, then use a [constant item](https://doc.rust-lang.org/reference/items/constant-items.html) as match arm:

```rust
const NULL_PTR: (some type) = null_mut();
match xs_handle {
  NULL_PTR => { ... }
  _ => { ... }
}
```

NB: I still cannot compile libmicrovmi, xenctrl, ... :( (but `vagrant` is ok now)

PR #2, opened 2019-06-17
Xen read physical

Wenzel:
Add incomplete implementation of `read_physical` API

uses `map`/`unmap` pages in `xenforeignmemory`:
https://github.com/Wenzel/xenforeignmemory/blob/master/src/lib.rs#L25

### DIFF
--- a/src/api.rs
+++ b/src/api.rs
@@ -4,6 +4,9 @@ pub enum DriverType {
 }
 
 pub trait Introspectable {
+    // read physical memory
+    fn read_physical(&self, paddr: u64, count: u32) -> Result<Vec<u8>,&str>;
+
     // pause the VM
     fn pause(&self);
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -7,6 +7,9 @@ pub trait Introspectable {
     // read physical memory
     fn read_physical(&self, paddr: u64, buf: &mut [u8]) -> Result<(),&str>;
 
+    // get max physical address
+    fn get_max_physical_addr(&self) -> Result<u64,&str>;
+
     // pause the VM
     fn pause(&self);
 
--- a/src/api.rs
+++ b/src/api.rs
@@ -5,7 +5,7 @@ pub enum DriverType {
 
 pub trait Introspectable {
     // read physical memory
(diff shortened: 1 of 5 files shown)


src/driver/xen.rs
```diff
@@ -49,6 +57,42 @@ impl Xen {
...
+            let map = self.xen_fgn.map(self.domid, PROT_READ, gfn).unwrap() as *mut c_char;
+            // determine how much we can read
+            if (offset + count as u64) > xenctrl::PAGE_SIZE as u64 {
+                read_len = (xenctrl::PAGE_SIZE as u64) - offset;
+            } else {
+                read_len = count_mut.into();
+            }
+
+            // convert to u8 array
+            let page = unsafe {
```
Wenzel: @tathanhdinh idealy, what we should do here is move `std::slice::from_raw_parts_mut` to the `map` call in `xenforeignmemory`, so that `map` returns an `&mut [u8]`.

However, how do we retrieve the pointer afterwards to unmap the page ?

src/api.rs
```diff
@@ -4,6 +4,9 @@ pub enum DriverType {
 }
 
 pub trait Introspectable {
+    // read physical memory
+    fn read_physical(&self, paddr: u64, count: u32) -> Result<Vec<u8>,&str>;
```
Wenzel: @tathanhdinh should this return a `Result<&mut [u8], &str>`, or a `Vec` ? I'm a bit lost here.

src/driver/xen.rs
```diff
@@ -49,6 +57,42 @@ impl Xen {
 
 impl api::Introspectable for Xen {
 
+    fn read_physical(&self, paddr: u64, count: u32) -> Result<Vec<u8>,&str> {
```
Wenzel: This implementation comes from libvmi:
https://github.com/libvmi/libvmi/blob/master/libvmi/read.c#L116

src/api.rs
Wenzel: Actually an array sounds better than a Vec, since we know the maximum size in advance, and it shouldn't not grow anyway.

PR #21, opened 2019-11-02
kvm: use get_maximum_gfn

Wenzel:
WIP

The API is buggy ATM
It seems that it returns the max gfn of the host...


rageagainsthepc:
This should be obsolete by now, right?

Wenzel:
Yes, you can close it.

PR #38, opened 2019-12-23
setup github actions for dummy

Wenzel:


Wenzel:
@rageagainsthepc I'm switching from Travis to Gitub Actions.

I find it has a better integration with Github, and jobs start faster.

Also, we can use some Actions from the Github marketplace to enhance our pull request reviews, like this Clippy one:
https://github.com/marketplace/actions/rust-clippy-check

I haven't found how to integrate it yet, will do that in another PR.

PR #43, opened 2020-01-02
add a C compatible interface

rageagainsthepc:


c_examples/libmicrovmi.h
```diff
@@ -0,0 +1,82 @@
+#ifndef LIBMICROVMI_H
```
Wenzel: We shouldn't include a generated file in the sources.
It will be outdated in a few commits, and the user should generate this file by himself.

c_examples/mem-dump.c
```diff
@@ -0,0 +1,47 @@
...
+
+void dump_memory(MicrovmiContext* driver, const char* vm_name) {
+    if (microvmi_pause(driver) == MicrovmiSuccess) {
+        printf("Paused.\n");
+    } else {
+        printf("Unable to pause VM.\n");
+        return;
+    }
+    uint64_t max_address;
+    if (microvmi_get_max_physical_addr(driver, &max_address) == MicrovmiSuccess) {
```
Wenzel: Curiously, I couldn't run this example.
I always had a max address of `0`:
Capture d’écran de 2020-01-19 22-47-43

c_examples/pause.c
```diff
@@ -0,0 +1,27 @@
+#include <stdio.h>
```
Wenzel: Maybe adding a configurable `sleep` time in between will make this example code more useful.

c_examples/regs-dump.c
```diff
@@ -0,0 +1,38 @@
...
+        printf("Unable to pause VM.\n");
+        return;
+    }
+    Registers regs;
+    memset(&regs, 0, sizeof(regs));
+    if (microvmi_read_registers(driver, 0, &regs) == MicrovmiSuccess) {
+        printf("rax: 0x%llx\n", regs.x86._0.rax);
+        printf("rbx: 0x%llx\n", regs.x86._0.rbx);
+        printf("rcx: 0x%llx\n", regs.x86._0.rcx);
+        printf("rdx: 0x%llx\n", regs.x86._0.rdx);
```
Wenzel: I believe we should also print `RIP`, `RSP`, `RBP`, `RSI`, `RDI` and the `RFLAGS` to show meaningful output (for 32 bits systems at least)

src/capi.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+#[allow(clippy::missing_safety_doc)]
+#[no_mangle]
+pub unsafe extern "C" fn microvmi_init(
+    domain_name: *const c_char,
+    driver_type: DriverType,
+) -> *mut MicrovmiContext {
+    let safe_domain_name = CStr::from_ptr(domain_name).to_string_lossy().into_owned();
+    let driver = init(&safe_domain_name, Some(driver_type.clone()));
```
Wenzel: We should be able to pass a NULL `driver_type` here, so the `init` will attempt each driver until there is a match.
That's the default behavior in [mem-dump.rs](https://github.com/Wenzel/libmicrovmi/blob/master/examples/mem-dump.rs#L27)

So the user doesn't have to edit the file to change the `driver_type`.

c_examples/mem-dump.c
rageagainsthepc: Did you initialize the driver with DriverType = None or 0? Because this is equal to initalizing the driver with `DriverType::Dummy` since enums are just ints in C and 0 is the first enum value.

rageagainsthepc: Gotta figure out a way to make this parameter optional...

c_examples/README.md
```diff
@@ -0,0 +1,22 @@
...
+To build the examples just use the makefile located in `c_examples`.
+It will also generate the header file for you provided you have installed cbindgen.
+You just have to make sure that you have already built *libmicrovmi*.
+
+## Executing the examples
+
+Command to execute an example as an unprivileged user:
+
+~~~
+sudo LD_LIBRARY_PATH="$LD_LIBRARY_PATH:../target/debug" <example> <vm_name>
```
Wenzel: `sudo` :thinking: 

If you are working with Xen, you need to be root, that's correct.
But in the general case, I would prefer to remove this indication, since it's not needed for KVM and VirtualBox.

src/api.rs
```diff
@@ -64,4 +67,6 @@ pub trait Introspectable {
     fn resume(&mut self) -> Result<(), Box<dyn Error>> {
         unimplemented!();
     }
+
+    fn get_driver_type(&self) -> DriverType;
```
Wenzel: @rageagainsthepc can you explain why you had to introduce a new API here, it's hard to follow for me ?
I would like to avoid changing the core of the Rust library just to provide a C compatibility layer.

rageagainsthepc: Well, if we initialize the driver with None, we don't know the type of the driver in advance. But we need to know since we are casting back and forth between void* and the actual type of the driver. As far as I know it's not possible to cast directly from void* to a trait type. There are probably different ways to approach this, but this one seemed pretty straight forward to me. I'm open to alternatives though ;)

rageagainsthepc: I think the issue was that pointers to traits are fat pointers. Maybe there is way to pass fat pointers to C code and cast them back to traits afterwards. I'll look into that.

rageagainsthepc: There exists a std::raw::TraitObject but it's marked as unstable because its representation may change in the future https://github.com/rust-lang/rust/issues/27751 . Also cbindgen reports TraitObject types as unsupported. In theory we could split the fat pointer into two separate pointers, store them in a struct and transmute the memory back to an Introspectable, but imho that sounds super ugly and unstable.

rageagainsthepc: On the other hand, we could just loose the support for passing `None` instead of the DriverType ;)

rageagainsthepc: I've found the crate downcast-rs. With this crate you are able to check a crate for its concrete type.

Wenzel: Okay, thanks for the update.

I'm okay to merge this as long as you add a comment that this API has been introduced for the solve purpose of the C interoperability, and that it's a workaround, to be deprecated in the future when new better solutions will emerge to deal with this, either from Rust unstable, or simply by switching to dynamic library loading.

PR #48, opened 2020-01-24
actions: test C API

Wenzel:
Following https://github.com/Wenzel/libmicrovmi/pull/43

We should add CI test before merging a PR in the future.

cc @rageagainsthepc

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -87,6 +87,23 @@ jobs:
     - name: build Hyper-V driver
       run: cargo build --features hyper-v
 
+  c_api:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-16.04, ubuntu-18.04]
+
+    steps:
+      - name: install cbindgen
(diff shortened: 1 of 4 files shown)


Wenzel:
@rageagainsthepc turns out I had to fix the Makefile ! :-)

The order of GCC arguments matter, especially the linking order, always put them at the end of the command line:
https://www.quora.com/Why-does-order-of-parameters-supplied-to-GCC-matter

rageagainsthepc:
Huh, it actually does ring a bell, but I've totally forgotten about it.

Wenzel:
@rageagainsthepc the funny thing is that it actually compiled in my Debian 10 VM :man_shrugging: 

Older version of GCC, different behavior I guess..

rageagainsthepc:
That can't be it. I've successfully compiled it on Fedora 31 with a very recent version of GCC. 🤔

Wenzel:
Well, on my ubuntu 19.10, it didn't.

Important thing is that we respect gcc's command line expectations.

PR #49, opened 2020-01-25
Kvm intercept cr3

Wenzel:
Started implementing CR3 interception for KVM.

### DIFF
--- a/examples/cr3-events.rs
+++ b/examples/cr3-events.rs
@@ -70,5 +70,10 @@ fn main() {
     drv.toggle_intercept(0, inter_cr3, false)
         .expect("Failed to enable CR3 interception");
 
-    println!("Catched {} events/sec", i / duration.as_secs());
+    println!(
+        "Catched {} events in {} seconds ({} events/sec)",
+        i,
+        duration.as_secs(),
+        i / duration.as_secs()
+    );
 }
--- a/examples/cr3-events.rs
+++ b/examples/cr3-events.rs
@@ -1,0 +1,44 @@
+use std::env;
+
+use env_logger;
+
+use microvmi::api::EventType;
+use microvmi::api::Introspectable;
+
+fn main() {
+    env_logger::init();
+
+    let args: Vec<String> = env::args().collect();
+    if args.len() != 2 {
+        println!("Usage: {} <vm_name>", args[0]);
+        return;
+    }
+    let domain_name = &args[1];
+
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+
+    println!("Enable CR3 interception");
+    drv.pause().expect("Failed to pause VM");
+
+    // enable CR3 interception
+    drv.toggle_intercept(0, EventType::CR3, true)
+        .expect("Failed to enable CR3 interception");
+
+    drv.resume().expect("Failed to resume VM");
+
+    println!("Listen for CR3 events...");
+    // listen
+    for i in 1..100 {
+        let event = drv.listen(1000);
+        println!("[{}] {:?}", i, event);
+    }
+
+    println!("Disable CR3 interception");
+    drv.pause().expect("Failed to pause VM");
+
+    // disable CR3 interception
+    drv.toggle_intercept(0, EventType::CR3, true)
+        .expect("Failed to enable CR3 interception");
+
+    drv.resume().expect("Failed to resume VM");
+}
--- a/examples/cr3-events.rs
+++ b/examples/cr3-events.rs
@@ -32,8 +32,10 @@ fn main() {
 
     // enable CR3 interception
     let inter_cr3 = InterceptType::Cr(CrType::Cr3);
-    drv.toggle_intercept(0, inter_cr3, true)
-        .expect("Failed to enable CR3 interception");
+    for vcpu in 0..drv.get_vcpu_count().unwrap() {
+        drv.toggle_intercept(vcpu, inter_cr3, true)
+            .expect("Failed to enable CR3 interception");
+    }
 
     drv.resume().expect("Failed to resume VM");
 
@@ -53,7 +55,7 @@ fn main() {
                         old: _,
                     } => new,
                 };
-                println!("[{}] VCPU {} - CR3: 0x{:x}", ev.vcpu, i, new);
+                println!("[{}] VCPU {} - CR3: 0x{:x}", i, ev.vcpu, new);
                 drv.reply_event(&ev, EventReplyType::Continue)
                     .expect("Failed to send event reply");
             }
@@ -67,8 +69,10 @@ fn main() {
     drv.pause().expect("Failed to pause VM");
 
     // disable CR3 interception
-    drv.toggle_intercept(0, inter_cr3, false)
-        .expect("Failed to enable CR3 interception");
+    for vcpu in 0..drv.get_vcpu_count().unwrap() {
+        drv.toggle_intercept(vcpu, inter_cr3, false)
+            .expect("Failed to enable CR3 interception");
+    }
 
     println!(
         "Catched {} events in {} seconds ({} events/sec)",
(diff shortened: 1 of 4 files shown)


Wenzel:
So, I implemented 2 new APIs:

~~~Rust
    // toggle an event interception
    fn toggle_intercept(
        &mut self,
        _vcpu: u16,
        _event_type: EventType,
        _enabled: bool,
    ) -> Result<(), Box<dyn Error>> {
        unimplemented!();
    }

    // listen and return the next event, or return None
    fn listen(&self, _timeout: u32) -> Result<Option<Event>, Box<dyn Error>> {
        unimplemented!();
    }
~~~

As well as a simple example to listen for the next `100` `CR3` changes.

But it doesn't work as expected:
Capture d’écran de 2020-01-25 15-15-56

I'm getting a segfault ATM, and I think I should have a second look at my `pop_event` implementation:
https://github.com/Wenzel/kvmi/blob/master/src/lib.rs#L215

@kylerky, if you have a few minutes, can you tell me if you see something absurd that i'm doing there, which would lead to a segfault ? (apart from not freeing the kvmi_event afterwards)

Thanks.

kylerky:
No idea. I suggest using Valgrind to look for memory corruptions.

I notice that pop_event succeeded when first called. Therefore, it may be that the bug is somewhere else.

Wenzel:
@kylerky I found the reason of the crash, I was dereferencing a null pointer, because I wasn't properly checking `kvmi_pop_event` 's return code:
fixed in https://github.com/Wenzel/kvmi/pull/12

The I discovered that the `kvmi` crate was not up to date with `kvmi-v6`, which requires new headers in event replies:
https://github.com/Wenzel/kvmi/pull/13

Now I'm stuck at designing an API to would let me set an event reply once I received it.
I added a new API to the trait:

~~~Rust
fn reply_event(&self, _event: Event, _reply_type: EventReplyType) -> Result<(), Box<dyn Error>> {
        unimplemented!();
}
~~~

Now I need to store the `kvmi_event` structure inside the generic libmicromi's `Event` struct.

But how can I do that ?
I needs to be generic enough so I can store a `XenEvent` or `VirtualBoxEvent`.

This is how I represent my Events right now:
~~~Rust
pub enum EventType {
    Cr { cr_type: CrType, new: u64, old: u64 },
}

pub struct Event {
    pub kind: EventType,
    driver_event: ????? // should be able to store a KVMiEvent, XenEvent or VirtualBoxEvent
}
~~~

Any ideas @kylerky ?

kylerky:
```rust
pub struct Event {
    common: EventCommon,
    extra: EventExtra,
}

pub enum EventExtra<CR>
    where CR: CREvent
{
    CR(CR),
    Pause,
}

pub trait CREvent {
    fn get_old() -> u64 {
        ...
    }
    fn get_new() -> u64 {
        ...
    }
}

pub struct KVMCrEvent {
    pub fn construct_reply(...) -> KVMCrReply {

    }
}

impl CREvent for KVMCrEvent {...}

struct KVMCrReply {...}
impl Reply for KVMCrReply {...}

fn main() {
    let kvm_cr_event = ...;
    let reply = kvm_cr_event.construct_reply(Continue, new);
    dom.send(reply);
}

fn send<T: Reply>(reply: T) { ... }

```

kylerky:
I think it is hard to fit `reply_event` into a single function because different types of event ask for different kinds of replies. We should at least go for a generic function.

In my proposal above, I go further as to having event as builders for event replies, and then we can use a generic `send` to send them.

Wenzel:
@kylerky , @rageagainsthepc

I finished a basic implementation of CR interception on KVM.
It doesn't handle modifying the CR values for now, we can deal with it in a separate PR (I'll create an issue for that and we propose ideas there).

src/driver/kvm.rs
```diff
@@ -106,13 +132,82 @@ impl Introspectable for Kvm {
...
+                };
+                Ok(self.kvmi.control_cr(vcpu, kvmi_cr, enabled)?)
+            }
+        }
+    }
+
+    fn listen(&mut self, timeout: u32) -> Result<Option<Event>, Box<dyn Error>> {
+        // wait for next event and pop it
+        debug!("wait for next event");
+        if self.kvmi.wait_event(timeout.try_into().unwrap())?.is_none() {
```
rageagainsthepc: nitpick: `kvmi.wait_event` returns `None` or an empty `Some()`. I would not expect Some() to yield no value. Maybe use a custom enum event/no_event instead?

rageagainsthepc: Also, I don't see why `wait_event` shouldn't return the event directly. Is there a special reason for not popping the event directly in `wait_event`?

Wenzel: I implemented the KVMi Rust API by doing a 1:1 mapping with the libkvmi functions.
But I suppose we could wrap these 2 calls into a single one `pop_event()` function.

src/driver/kvm.rs
```diff
@@ -106,13 +132,82 @@ impl Introspectable for Kvm {
...
+            KVMiEventType::Cr { cr_type, new, old } => EventType::Cr {
+                cr_type: match cr_type {
+                    KVMiCr::Cr0 => CrType::Cr0,
+                    KVMiCr::Cr3 => CrType::Cr3,
+                    KVMiCr::Cr4 => CrType::Cr4,
+                },
+                new,
+                old,
+            },
+            KVMiEventType::PauseVCPU => panic!("Unexpected PauseVCPU event. It should have been poped by resume VM. (Did you forgot to resume your VM ?)"),
```
rageagainsthepc: nitpick: *popped, *forget

src/driver/kvm.rs
```diff
@@ -1,32 +1,58 @@
...
+    X86Registers,
+};
 
 // unit struct
 #[derive(Debug)]
 pub struct Kvm {
     kvmi: KVMi,
     expect_pause_ev: u32,
+    // VCPU -> KVMiEvent
+    map_events: HashMap<u16, KVMiEvent>,
```
rageagainsthepc: nitpick: From what I understand you want to register a single event per vcpu, right? Isn't a hash map a little bit of an overkill for this purpose? I would just use a vector instead. You don't necessarily need to change this, I mainly want to know if I have missed something. If so, please tell me ;)

src/driver/kvm.rs
rageagainsthepc: Either way works. :+1:

src/api.rs
```diff
@@ -71,7 +76,63 @@ pub trait Introspectable {
...
+    }
+
+    // listen and return the next event, or return None
+    fn listen(&mut self, _timeout: u32) -> Result<Option<Event>, Box<dyn Error>> {
+        unimplemented!();
+    }
+
+    fn reply_event(
+        &mut self,
+        _event: &Event,
```
rageagainsthepc: Just a thought: It would make sense to consume the event here, right? This would automatically make sure that the user replies only once per event. Now we just need to ensure that the user sends a reply at all. ;)

src/driver/kvm.rs
Wenzel: Yes, it's totally overkill.

I just wanted to store the KVMEvent in the KVM driver, and associate its VCPU id to find it back for the response.

A vector is good also.
 
And i'm not storing singlestep events but general KVMevents.

examples/cr3-events.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    // disable CR3 interception
+    for vcpu in 0..drv.get_vcpu_count().unwrap() {
+        drv.toggle_intercept(vcpu, inter_cr3, false)
+            .expect("Failed to enable CR3 interception");
+    }
+
+    let ev_per_sec = i as f64 / duration.as_secs_f64();
+    println!(
+        "Catched {} events in {:.2} seconds ({:.2} events/sec)",
```
rageagainsthepc: nitpick: *Caught

src/driver/kvm.rs
Wenzel: The last version uses a Vector instead.

Wenzel:
@rageagainsthepc merging this one.

PR #52, opened 2020-01-27
Fix kvmi_event pattern matching

rageagainsthepc:


Wenzel:
Why the CI tests didn't triggered ? :thinking: 
I can't see them

PR #53, opened 2020-01-27
actions: trigger build on PR

Wenzel:
This should fix https://github.com/Wenzel/libmicrovmi/pull/52 not triggering a CI build.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push]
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -1,6 +1,6 @@
 name: Build
 
-on: [push]
+on: [pull_request]
```
rageagainsthepc: It would be nice to have both, wouldn't it?

PR #58, opened 2020-02-13
Kvm intercept cr3 generics

Wenzel:
I tried my own implementation using generics and associated types.

The goal was to hide a generic `T` in Event<T>, so that It can contain a `KVMIEvent`, `XenEvent`, etc... in order to get the same in `reply_event()`

Unfortunately it doesn't compile.
@kylerky can I ask to check why ?

I'm puzzled by traits and generics here.


Wenzel:
This was a bad idea

PR #70, opened 2020-04-25
Driver init option

Wenzel:
This PR adds support for additional driver initialization parameters.

This is required for KVM driver, which needs a socket parameter.
Before that it was simply hardcoded to `/tmp/introspector` in the driver.

I added documentation for the new type `DriverInitParam`.
and I converted all examples to the `clap` argument parsing library, so it's easier to support command line parameters in the future.

Naming: what is better:

- `DriverInitParam`
- `DriverInitOption`
- other naming suggestions
:thinking:

### DIFF
--- a/examples/cr-events.rs
+++ b/examples/cr-events.rs
@@ -6,7 +6,9 @@ use clap::{App, Arg, ArgMatches};
 use colored::*;
 use env_logger;
 
-use microvmi::api::{CrType, EventReplyType, EventType, InterceptType, Introspectable};
+use microvmi::api::{
+    CrType, DriverInitParam, EventReplyType, EventType, InterceptType, Introspectable,
+};
 
 fn parse_args() -> ArgMatches<'static> {
     App::new(file!())
@@ -21,6 +23,14 @@ fn parse_args() -> ArgMatches<'static> {
                 .short("r")
                 .default_value("3")
                 .help("control register to intercept. Possible values: [0 3 4]"),
+        )
+        .arg(
+            Arg::with_name("kvmi_socket")
+                .short("k")
+                .takes_value(true)
+                .help(
+                "pass additional KVMi socket initialization parameter required for the KVM driver",
+            ),
         )
         .get_matches()
 }
@@ -77,7 +87,10 @@ fn main() {
     .expect("Error setting Ctrl-C handler");
 
     println!("Initialize Libmicrovmi");
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = matches
+        .value_of("kvmi_socket")
+        .map(|socket| DriverInitParam::KVMiSocket(socket.into()));
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None, init_option);
 
     // enable control register interception
     toggle_cr_intercepts(&mut drv, &vec_cr, true);
--- a/examples/interrupt-events.rs
+++ b/examples/interrupt-events.rs
@@ -6,13 +6,21 @@ use clap::{App, Arg, ArgMatches};
 use colored::*;
 use env_logger;
 
-use microvmi::api::*;
+use microvmi::api::{DriverInitParam, EventReplyType, EventType, InterceptType, Introspectable};
 
 fn parse_args() -> ArgMatches<'static> {
     App::new(file!())
         .version("0.1")
         .about("Watches interrupt VMI events")
         .arg(Arg::with_name("vm_name").index(1).required(true))
+        .arg(
+            Arg::with_name("kvmi_socket")
+                .short("k")
+                .takes_value(true)
+                .help(
+                "pass additional KVMi socket initialization parameter required for the KVM driver",
+            ),
+        )
         .get_matches()
 }
 
@@ -37,6 +45,9 @@ fn main() {
 
     let domain_name = matches.value_of("vm_name").unwrap();
 
+    let init_option = matches
+        .value_of("kvmi_socket")
+        .map(|socket| DriverInitParam::KVMiSocket(socket.into()));
     // set CTRL-C handler
     let running = Arc::new(AtomicBool::new(true));
     let r = running.clone();
@@ -46,7 +57,7 @@ fn main() {
     .expect("Error setting Ctrl-C handler");
 
     println!("Initialize Libmicrovmi");
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None, init_option);
 
     //Enable int3 interception
     toggle_int3_interception(&mut drv, true);
--- a/examples/mem-dump.rs
+++ b/examples/mem-dump.rs
@@ -1,30 +1,45 @@
-use std::env;
 use std::fs::File;
 use std::io::Write;
 use std::path::Path;
 
-extern crate microvmi;
+use clap::{App, Arg, ArgMatches};
 use env_logger;
 
-// traits method can only be used if the trait is in the scope
-use microvmi::api::Introspectable;
+use microvmi::api::{DriverInitParam, Introspectable};
 
 const PAGE_SIZE: usize = 4096;
+
+fn parse_args() -> ArgMatches<'static> {
+    App::new(file!())
+        .version("0.1")
+        .author("Mathieu Tarral")
+        .about("Dumps VM physical memory")
+        .arg(Arg::with_name("vm_name").index(1).required(true))
+        .arg(
+            Arg::with_name("kvmi_socket")
+                .short("k")
+                .takes_value(true)
+                .help(
+                "pass additional KVMi socket initialization parameter required for the KVM driver",
+            ),
+        )
+        .get_matches()
+}
 
 fn main() {
     env_logger::init();
 
-    let args: Vec<String> = env::args().collect();
-    if args.len() != 2 {
-        println!("Usage: {} <vm_name>", args[0]);
-        return;
-    }
-    let domain_name = &args[1];
+    let matches = parse_args();
+    let domain_name = matches.value_of("vm_name").unwrap();
+
     let dump_name = format!("{}.dump", domain_name);
     let dump_path = Path::new(&dump_name);
     let mut dump_file = File::create(dump_path).expect("Fail to open dump file");
 
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = matches
+        .value_of("kvmi_socket")
+        .map(|socket| DriverInitParam::KVMiSocket(socket.into()));
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None, init_option);
 
     println!("pausing the VM");
     drv.pause().expect("Failed to pause VM");
--- a/examples/msr-events.rs
+++ b/examples/msr-events.rs
@@ -6,7 +6,7 @@ use std::sync::Arc;
 use std::time::Instant;
 use std::u32;
 
-use microvmi::api::{EventReplyType, EventType, InterceptType, Introspectable};
+use microvmi::api::{DriverInitParam, EventReplyType, EventType, InterceptType, Introspectable};
 
 // default set of MSRs to be intercepted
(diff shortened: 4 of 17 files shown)


Wenzel:
I tried to adapt the capi, but the generated libmicrovmi.h doesn't compile.
You can see the branch https://github.com/Wenzel/libmicrovmi/tree/driver_init_option_capi

~~~
gcc mem-dump.c -o mem-dump -lmicrovmi -L../target/debug
In file included from mem-dump.c:5:
libmicrovmi.h:51:12: error: field ‘_0’ has incomplete type
   51 |     String _0;
      |            ^~
make: *** [Makefile:14: mem-dump] Error 1
~~~

~~~C
typedef enum {
    KVMiSocket,
} DriverInitParam_Tag;

typedef struct {
    String _0;
} KVMiSocket_Body;

typedef struct {
    DriverInitParam_Tag tag;
    union {
        KVMiSocket_Body kvmi_socket;
    };
} DriverInitParam;
~~~

rageagainsthepc:
I think that you get this error because there is no representation of a rust `String` in C.

Wenzel:
But it's it cbindgen's job to replace this String type with something manageable in C ?

rageagainsthepc:
BIndgen doesn't do that. That's why you need the `#[repr(C)]` for every struct or enum. I would create an extra enum for ffi that has got a CStr instead of a rust string inside. Then I would implement the `Into` to trait for it so you can easily convert it into you internal representation.

src/api.rs
```diff
@@ -14,6 +15,42 @@ pub enum DriverType {
...
+/// ```
+///let kvmi_socket = DriverInitParam::KVMiSocket(String::from("/tmp/introspector"));
+///let mut drv = microvmi::init(domain_name, DriverType::KVM, Some(kvmi_socket));
+/// ```
+#[derive(Debug)]
+pub enum DriverInitParam {
+    KVMiSocket(String),
+}
+
+impl From<DriverInitParamFFI> for DriverInitParam {
```
Wenzel, line 38: @rageagainsthepc I tried to follow your advice and added another enum with a CString instead, and a From trait implementation (should I use Into) ?

It doesn't compile right now because there is a move, and I don't know how to solve it
Capture d’écran de 2020-08-09 17-16-29

if you have any idea.

Wenzel:
@rageagainsthepc 
~~~
 gcc mem-dump.c -o mem-dump -lmicrovmi -L../target/debug
In file included from mem-dump.c:5:0:
libmicrovmi.h:31:5: error: unknown type name ‘CString’
     CString _0;
     ^~~~~~~
~~~

even a `CString` isn't translated.

Wenzel:
Finally, a green CI on this old PR.
@rageagainsthepc , do you want to review it ?

rageagainsthepc:
@Wenzel I do.

src/api.rs
rageagainsthepc: `From` is usually better than `Into` because you get the `Into` implementation for free IIRC. However since for converion is fallible you should perhaps consider using `TryFrom` instead.

src/driver/hyperv.rs
```diff
@@ -6,7 +6,7 @@ use std::ptr::{null, null_mut};
 use std::slice;
 use std::vec::Vec;
 
-use crate::api::Introspectable;
+use crate::api::{DriverInitParam, DriverType, Introspectable};
```
rageagainsthepc, line 9: ... What the CI says ;)

src/driver/kvm.rs
```diff
@@ -320,7 +325,13 @@ mod tests {
             ))
         });
 
-        let result = Kvm::new("some_vm", kvmi_mock);
+        let result = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket(
+                "/tmp/introspector".parse().unwrap(),
+            )),
```
rageagainsthepc, line 333:
```suggestion
            Some(DriverInitParam::KVMiSocket("/tmp/introspector")),
```

src/driver/kvm.rs
```diff
@@ -391,7 +402,13 @@ mod tests {
                 .returning(|_, _, _| Ok(()));
         }
 
-        let result = Kvm::new("some_vm", kvmi_mock);
+        let result = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket(
+                "/tmp/introspector".parse().unwrap(),
+            )),
```
rageagainsthepc, line 410:
```suggestion
            Some(DriverInitParam::KVMiSocket("/tmp/introspector")),
```

rageagainsthepc:
You have added the possibility to pass the socket path in all rust examples. Why not do the same with the c examples

examples/cr-events.rs
```diff
@@ -77,7 +87,11 @@ fn main() {
     .expect("Error setting Ctrl-C handler");
 
     println!("Initialize Libmicrovmi");
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = match matches.value_of("kvmi_socket") {
+        None => None,
+        Some(socket) => Some(DriverInitParam::KVMiSocket(String::from(socket))),
+    };
```
rageagainsthepc, line 93:
```suggestion
    let init_option = matches
        .value_of("kvmi_socket")
        .map(|socket| DriverInitParam::KVMiSocket(String::from(socket)));
```

examples/interrupt-events.rs
```diff
@@ -37,6 +45,10 @@ fn main() {
 
     let domain_name = matches.value_of("vm_name").unwrap();
 
+    let init_option = match matches.value_of("kvmi_socket") {
+        None => None,
+        Some(socket) => Some(DriverInitParam::KVMiSocket(String::from(socket))),
+    };
```
rageagainsthepc, line 51:
```suggestion
    let init_option = matches
        .value_of("kvmi_socket")
        .map(|socket| DriverInitParam::KVMiSocket(String::from(socket)));
```

examples/mem-dump.rs
```diff
@@ -1,30 +1,46 @@
...
-    let domain_name = &args[1];
+    let matches = parse_args();
+    let domain_name = matches.value_of("vm_name").unwrap();
+
     let dump_name = format!("{}.dump", domain_name);
     let dump_path = Path::new(&dump_name);
     let mut dump_file = File::create(dump_path).expect("Fail to open dump file");
 
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = match matches.value_of("kvmi_socket") {
```
rageagainsthepc, line 39: Same here.

examples/msr-events.rs
```diff
@@ -75,6 +83,11 @@ fn main() {
         registers = DEFAULT_MSR.to_vec();
     }
 
+    let init_option = match matches.value_of("kvmi_socket") {
```
rageagainsthepc, line 86: Same here.

examples/pause.rs
```diff
@@ -1,24 +1,45 @@
...
-        return;
-    }
-    let domain_name = &args[1];
-    let timeout = args[2].parse::<u64>().unwrap();
+    let matches = parse_args();
+    let domain_name = matches.value_of("vm_name").unwrap();
+    let timeout = matches.value_of("timeout").unwrap().parse::<u64>().unwrap();
 
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = match matches.value_of("kvmi_socket") {
```
rageagainsthepc, line 38: Same here.

examples/regs-dump.rs
```diff
@@ -1,21 +1,36 @@
...
-    if args.len() != 2 {
-        println!("Usage: {} <vm_name>", args[0]);
-        return;
-    }
-    let domain_name = &args[1];
+    let matches = parse_args();
+    let domain_name = matches.value_of("vm_name").unwrap();
 
-    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    let init_option = match matches.value_of("kvmi_socket") {
```
rageagainsthepc, line 29: Same here. There sure are a lot of examples.

Wenzel:
> You have added the possibility to pass the socket path in all rust examples. Why not do the same with the c examples?
Because Rust has clap for argument parsing, and C has nothing, so it's always tedius to add support for another command line argument, and I just passed.

src/driver/kvm.rs
Wenzel: Capture d’écran de 2020-08-10 21-00-40
it's weird, because the code compiles, but my IDE suggests it's an error

Wenzel:
Thanks for your review @rageagainsthepc !
I just pushed the changes

src/driver/kvm.rs
rageagainsthepc: It do be like that sometimes with the intellij rust plugin ;)

rageagainsthepc:
> Because Rust has clap for argument parsing, and C has nothing, so it's always tedius to add support for another command line argument, and I just passed.

Yeah I totally get that. But why not do a low effort version where you just check whether `argc > 2` and if it is true you pass `argv[2]` to the init function. So, no switches and whatnot, just an optional second param.

rageagainsthepc:
But I am also fine with it if you just merge the current state, your call.

Wenzel:
If i add argument parsing in the C examples, it will a solid thing like getopt

src/api.rs
```diff
@@ -14,6 +17,42 @@ pub enum DriverType {
...
+            ),
+        })
+    }
+}
+
+/// Support passing initialization options
+/// similar to DriverInitParam, however this enum offers C API compatibility
+#[repr(C)]
+#[derive(Debug)]
+pub enum DriverInitParamFFI {
```
rageagainsthepc, line 52: This should be part of the capi, not of the regular api imho.

src/api.rs
```diff
@@ -14,6 +17,42 @@ pub enum DriverType {
...
+    fn try_from(param: DriverInitParamFFI) -> Result<Self, Self::Error> {
+        Ok(match param {
+            DriverInitParamFFI::KVMiSocket(cstr_socket) => DriverInitParam::KVMiSocket(
+                unsafe { CStr::from_ptr(cstr_socket) }
+                    .to_owned()
+                    .into_string()?,
+            ),
+        })
+    }
+}
```
rageagainsthepc, line 46: I kinda gave weird advice to be honest. What I said about the `From` trait does not apply to the `TryFrom` trait as there is no blanket implementation for `TryInto`. So conversely to `From` you do not get `TryInto` for free. Therefore I would suggest:
```suggestion
impl TryInto<DriverInitParam> for DriverInitParamFFI {
    type Error = IntoStringError;

    fn try_into(self) -> Result<DriverInitParam, Self::Error> {
        Ok(match self {
            DriverInitParamFFI::KVMiSocket(cstr_socket) => DriverInitParam::KVMiSocket(
                unsafe { CStr::from_ptr(cstr_socket) }
                    .to_owned()
                    .into_string()?,
            ),
        })
    }
}
```
And I would also move it to the capi.

src/api.rs
Wenzel: Fixed !

Wenzel:
I still don't get why we have a failed CI.
I configured the threshold to allow 10% drop and be a success.

Wenzel:
@rageagainsthepc I don't get it.
I tried to:
- increase the drop threshold to 10%
- force the if_ci_failed flag in codecode.yaml root file
- rename codecov.yaml to codecov.yml

nothing works, still fails.

Can we just remove the upload to codecov ?
This makes me loose my time and prevent me from merging PRs at this point, just like in https://github.com/ros-planning/moveit/issues/1896

rageagainsthepc:
Let us try this one more thing with the `informational: true` flag and if that does not help, fuck it.

Wenzel:
Can you push on this branch, just for testing ?

rageagainsthepc:
Sure

Wenzel:
Ah, I thought you couldn't push so I did it also :p

rageagainsthepc:
I see xD You gave me write access a few weeks ago, member? I think you had an error in your codecov.yml btw. It should be `project`, not `patch`.

Wenzel:
Finally, a green CI :green_circle: 
Can you make the same changes in a separate PR ?
I will rebase on this one.
Thanks.

rageagainsthepc:
Sure thing.

PR #74, opened 2020-06-04
Ornib summer

arnabcs17b006:


src/driver/kvm.rs
```diff
@@ -55,16 +55,19 @@ impl Introspectable for Kvm {
...
+        Ok(self.kvmi.write_physical(paddr, buf)?)
+    }
+
     fn get_max_physical_addr(&self) -> Result<u64, Box<dyn Error>> {
-        // No API in KVMi at the moment
-        // fake 512MB
-        let max_addr = 1024 * 1024 * 512;
-        Ok(max_addr)
+        let max_gfn = self.kvmi.get_maximum_gfn()?;
+        Ok(max_gfn << 12)
```
rageagainsthepc, line 64: nitpick: Try to avoid magic values. Better define a constant for that.

src/driver/kvm.rs
```diff
@@ -84,13 +87,88 @@ impl Introspectable for Kvm {
...
+                limit: _sregs.ldt.limit,
+                selector: _sregs.ldt.selector,
+            },
         }))
     }
 
+    fn write_registers(&self, vcpu: u16, value: u64, reg: Register) -> Result<(), Box<dyn Error>> {
+        let (mut regs, _sregs, _msrs) = self.kvmi.get_registers(vcpu)?;
+        match reg {
+            a if a == Register::RAX => regs.rax = value,
```
rageagainsthepc, line 148: `Register::RAX => regs.rax = value,` would be better

rageagainsthepc: Also, an even cleaner way to perform conversion would be to use the from or into trait. Have a look at https://doc.rust-lang.org/stable/rust-by-example/conversion/from_into.html

src/driver/kvm.rs
```diff
@@ -84,13 +87,88 @@ impl Introspectable for Kvm {
...
+            ldt: segment_reg {
+                base: _sregs.ldt.base,
+                limit: _sregs.ldt.limit,
+                selector: _sregs.ldt.selector,
+            },
         }))
     }
 
+    fn write_registers(&self, vcpu: u16, value: u64, reg: Register) -> Result<(), Box<dyn Error>> {
+        let (mut regs, _sregs, _msrs) = self.kvmi.get_registers(vcpu)?;
```
rageagainsthepc, line 146: I would consider removing register fetching here and instead expecting a whole register struct. This way you move the responsibility of calling get_registers up to the user of the API which makes it easier to understand what's going on, imho. Plus, when the user needs to call get_registers anyway prior to setting values you end up having redundant function calls.

arnabcs17b006:
Sure I will do that. But won't it break the abstraction? Also I think we
generally do not expose the user to kvmi apis
(as per the trends in other examples present)

arnabcs17b006:
Sorry. My mistake. I did not know the conventions.

rageagainsthepc:
> Sure I will do that. But won't it break the abstraction? Also I think we generally do not expose the user to kvmi apis (as per the trends in other examples present)

Of course you should not pass the kvm register structs through to the user, that's not what I meant ;) But you could convert those to a unified register representation in libmicrovmi. Let's see what @Wenzel thinks about that.

rageagainsthepc:
> Sir I think I need the guard variable a here, otherwise it will match in the first case itself and exit out of the match structure

No, that's the canonic way of matching enums. The code might have been useful this way when you where still matching plain integers, but you are not doing that anymore here. Try it out for yourself.

arnabcs17b006:
Sure

Wenzel:
Just a few things regarding your commit history:
- you have 2 typos in your commit
- you have commit names about temporary development state like 'draft' that could be squashed.
- you are changing the URL of `kvmi` crate. Better make a another PR from your fork of `kvmi` to the main repo, and keep the same URL.

BTW we should publish these crates to crates.io, and stop using URL.
Also using the Github CI to automatically publish a new version to crates.io whenever a branch is merged on master is a good thing to implement.

Cargo.toml
```diff
@@ -17,23 +17,32 @@ virtualbox = ["fdp"]
...
+enum-primitive-derive = "0.1.2"
+mockall = "0.7.1"
+kvmi-sys = { git = "https://github.com/Wenzel/kvmi-sys.git" }
 
 [dev-dependencies]
 env_logger = "0.7.1"
 ctrlc = "3.1.3"
 clap = "2.33.0"
 colored = "1.9.3"
+
```
rageagainsthepc, line 42: A side note: Throughout your code I've seen several occasions where you insert or delete lines without a specific reason. It would be nice if you avoided such changes. In this particular case it helps to active the setting "Trim Final Newlines" in vscode for example.

Wenzel:
@arnabcs17b006 , this PR is becoming messy, as you are adding multiple features at the sime time, and keep iterating every week.

I would like you to split it into multiple branches, one for each feature, or week, so the review is easier for us.

Second thing: you commited an empty `winxp.dump` file. please rewrite your commits and get rid of it.

examples/mem-events.rs
```diff
@@ -0,0 +1,104 @@
...
+
+    println!("Initialize Libmicrovmi");
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+    println!("Listen for memory events...");
+    // record elapsed time
+    let start = Instant::now();
+    get_pagefaults(&mut drv, true);
+    let mut i: u64 = 0;
+
+    //Code snippet to get page fault
```
Wenzel, line 60: @arnabcs17b006 to trigger some pagefaults in KVM, you need
1. enable PF events
2. remove a permission access (r/w/x) on a page which needs it

Here I can see that you do 1. , but then you do:

- curr_addr = 0
- var = PageAccessR
- buffer = [0; 4096]
- get_page_access(0)
- set_page_access(0, new accesses)
- read_physical(0)

There is nothing happened at address 0.
it's a page that's marked not readable, not executable, and not writable by the OS, in case of a null pointer dereference (I believe).

the problem of libkvmi is that i needs physical addresses to work with.
this example in Libvmi uses virtual address translation to find back the physical address of the current RIP, get the frame, and remove the execute permissions, triggering some pagefaults:
https://github.com/KVM-VMI/libvmi/blob/kvmi/examples/mem-event-example.c

Maybe you could iterate get the max gfn, and iterate over every frame to remove the execute permissions, over the entire RAM ?
What do you think @rageagainsthepc  ?

examples/cr-events.rs
```diff
@@ -20,7 +20,7 @@ fn parse_args() -> ArgMatches<'static> {
                 .takes_value(true)
                 .short("r")
                 .default_value("3")
-                .help("control register to intercept. Possible values: [0 3 4]"),
+                .help("control register to intercept. Possible values: [0 2 3 4]"),
```
Wenzel, line 23: CR2 is not a interceptable control register.

Please remove it.

examples/cr-events.rs
```diff
@@ -93,9 +93,11 @@ fn main() {
                         new,
                         old: _,
                     } => (cr_type, new),
+                    _ => (CrType::Cr3, 0),
                 };
                 let cr_color = match cr_type {
                     CrType::Cr0 => "blue",
+                    CrType::Cr2 => "black",
```
Wenzel, line 100: Same here

examples/cr-events.rs
```diff
@@ -93,9 +93,11 @@ fn main() {
                         new,
                         old: _,
                     } => (cr_type, new),
+                    _ => (CrType::Cr3, 0),
```
Wenzel, line 96: Why do you match `_` to CR3 here ?
this means that anything that is ot Cr EventType will be considered a CR3 event ???

examples/mem-events.rs
Wenzel: Actually we can mimik how `kvmi-test.c` does:
https://github.com/KVM-VMI/kvm/blob/kvmi/tools/kvm/kvmi/kvmi-test.c#L207

They change the permission of the current process's page tables, to get some events

Wenzel:
@arnabcs17b006 is this PR still relevant ?
otherwise you can close it

PR #76, opened 2020-06-19
Use actions rs

rageagainsthepc:
I have to admit, I am not really willing to deal with the clippy lints for hyper-v. Should I remove the check then or better just don't let it fail?


Wenzel:
You had clippy warnings for hyper-V driver ?
I guess clippy got new warnings since I commited this code.
We can skip this hyper-v clippy check, as there is no work on hyper-v driver since a few months.

PR #77, opened 2020-06-20
add kvm driver tests

rageagainsthepc:


rageagainsthepc:
@Wenzel I am not really sure I made the right decision. Normally you would wrap external dependencies and mock the wrapper, but I wanted to avoid another layer of indirection. However this would be cleaner from a design point of view and it would also result in less boiler plate code. So what do you think?

rageagainsthepc:
I prefer this approach right now because it introduces no overhead. Still have to rebase my kvmi branch on master though and this is where it gets tricky because libmicrovmi does not build right now with the latest revision of kvmi

src/driver/kvm.rs
```diff
@@ -165,6 +167,7 @@ impl Introspectable for Kvm {
                         old,
                     },
                     KVMiEventType::PauseVCPU => panic!("Unexpected PauseVCPU event. It should have been popped by resume VM. (Did you forget to resume your VM ?)"),
+                    _ => unimplemented!()
```
rageagainsthepc, line 170: Quick fix so everything compiles neatly. This should be removed anyways as soon as we are starting to merge @arnabcs17b006's PRs

Cargo.toml
```diff
@@ -37,3 +37,8 @@ env_logger = "0.7.1"
 ctrlc = "3.1.3"
 clap = "2.33.0"
 colored = "1.9.3"
+mockall = "0.7.1"
+test-case = "1.0.0"
+
+[patch."https://github.com/Wenzel/kvmi"]
+kvmi = { git = "https://github.com/rageagainsthepc/kvmi", branch = "testing-poc" }
```
rageagainsthepc, line 44: note to self: remove this as soon as we merge this branch into kvmi master.

rageagainsthepc:
@Wenzel Oh no, now you have merged this without giving me time to remove my patch from Cargo.toml

PR #80, opened 2020-06-22
Read/write physical memory

arnabcs17b006:
Do not merge until previous ones are merged!

### DIFF
--- a/c_examples/mem-dump.c
+++ b/c_examples/mem-dump.c
@@ -5,7 +5,6 @@
 #include "libmicrovmi.h"
 
 
-size_t PAGE_SIZE = 4096;
 
 void dump_memory(void* driver, const char* vm_name) {
     if (microvmi_pause(driver) == MicrovmiSuccess) {
--- a/examples/regs-dump.rs
+++ b/examples/regs-dump.rs
@@ -3,7 +3,7 @@ extern crate env_logger;
 extern crate microvmi;
 
 // traits method can only be used if the trait is in the scope
-use microvmi::api::{Introspectable, Registers};
+use microvmi::api::Introspectable;
 
 fn main() {
     env_logger::init();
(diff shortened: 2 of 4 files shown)


examples/cr-events.rs
```diff
@@ -20,7 +20,7 @@ fn parse_args() -> ArgMatches<'static> {
                 .takes_value(true)
                 .short("r")
                 .default_value("3")
-                .help("control register to intercept. Possible values: [0 3 4]"),
+                .help("control register to intercept. Possible values: [0 2 3 4]"),
```
Wenzel, line 23: same thing here, remove CR2 register

examples/cr-events.rs
```diff
@@ -93,9 +93,11 @@ fn main() {
                         new,
                         old: _,
                     } => (cr_type, new),
+                    _ => (CrType::Cr3, 0),
```
Wenzel, line 96: This doesn't make sense.
If the event is not a Cr, we should do some error handling, and not convert it to fake CR3

examples/msr-events.rs
```diff
@@ -0,0 +1,95 @@
...
+    let start = Instant::now();
+    // listen
+    let mut i: u64 = 0;
+    while running.load(Ordering::SeqCst) {
+        let event = drv.listen(1000).expect("Failed to listen for events");
+        match event {
+            Some(ev) => {
+                let (msr_type, new, old) = match ev.kind {
+                    EventType::Msr { msr_type, new, old } => (msr_type, new, old),
+                    _ => unimplemented!(),
```
rageagainsthepc, line 65: You should use `panic!` instead of `unimplemented!` here. It's not like you haven't implemented the functionality yet, but rather you're not expecting any other events to occur.

rageagainsthepc:
1. I would like you to open a separate PR for your msr intercept functionality. It's not nice to review two features in one commit which really ought to be separate from each other.
2. There is a lot of code you have duplicated for your msr-event example. You should consider simply adding another switch to the cr intercept example (of course while also renaming it in the process) which let's you choose the msr registers you want to intercept.

examples/cr-events.rs
```diff
@@ -89,6 +89,7 @@ fn main() {
             Some(ev) => {
                 let (cr_type, new, old) = match ev.kind {
                     EventType::Cr { cr_type, new, old } => (cr_type, new, old),
+                    _ => panic!("not cr event"),
```
rageagainsthepc, line 92: This line belongs to the msr PR as well afais.

PR #82, opened 2020-06-22
Unit tests

arnabcs17b006:
Do not merge until previous ones are merged!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,16 +24,21 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/arnabcs17b006/kvmi", optional = true }
 fdp = { git = "https://github.com/Wenzel/fdp", optional = true }
+kvmi = { git = "https://github.com/arnabcs17b006/kvmi", optional = true }
 winapi = { version = "0.3.8", features = ["tlhelp32", "winnt", "handleapi", "securitybaseapi"], optional = true }
 widestring = { version = "0.4.0", optional = true }
 ntapi = { version = "0.3.3", optional = true }
 vid-sys = { version = "0.3.0", features = ["deprecated-apis"], optional = true }
 cty = "0.2.1"
+enum-primitive-derive = "0.1.2"
 
 [dev-dependencies]
 env_logger = "0.7.1"
 ctrlc = "3.1.3"
 clap = "2.33.0"
 colored = "1.9.3"
+
+
+
+
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", optional = true }
+kvmi = { git = "https://github.com/arnabcs17b006/kvmi", optional = true }
 fdp = { git = "https://github.com/Wenzel/fdp", optional = true }
 winapi = { version = "0.3.8", features = ["tlhelp32", "winnt", "handleapi", "securitybaseapi"], optional = true }
 widestring = { version = "0.4.0", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ virtualbox = ["fdp"]
 # Hyper-v driver
 hyper-v = ["winapi", "widestring", "ntapi", "vid-sys"]
 
-
 [dependencies]
 log = "0.4.8"
 libc = { version = "0.2.58", optional = true }
@@ -32,12 +31,17 @@ ntapi = { version = "0.3.3", optional = true }
 vid-sys = { version = "0.3.0", features = ["deprecated-apis"], optional = true }
 cty = "0.2.1"
 enum-primitive-derive = "0.1.2"
+mockall = "0.7.1"
+kvmi-sys = { git = "https://github.com/Wenzel/kvmi-sys.git" }
 
 [dev-dependencies]
 env_logger = "0.7.1"
 ctrlc = "3.1.3"
 clap = "2.33.0"
 colored = "1.9.3"
+
+
+
 
 
 
--- a/src/driver/kvm.rs
+++ b/src/driver/kvm.rs
@@ -5,10 +5,7 @@ use std::vec::Vec;
 
 use kvmi::{KVMi, KVMiCr, KVMiEvent, KVMiEventReply, KVMiEventType, KVMiInterceptType};
 
-use crate::api::{
-    CrType, DriverType, Event, EventReplyType, EventType, InterceptType, Introspectable, Registers,
-    X86Registers,
-};
+use crate::api::*;
 
 // unit struct
 #[derive(Debug)]
@@ -65,7 +62,8 @@ impl Introspectable for Kvm {
     fn read_registers(&self, vcpu: u16) -> Result<Registers, Box<dyn Error>> {
         let (regs, sregs, _msrs) = self.kvmi.get_registers(vcpu)?;
         // TODO: hardcoded for x86 for now
-        Ok(Registers::X86(X86Registers {
+
+        Ok((Registers::X86(X86Registers {
             rax: regs.rax,
             rbx: regs.rbx,
             rcx: regs.rcx,
@@ -85,10 +83,90 @@ impl Introspectable for Kvm {
             rip: regs.rip,
             rflags: regs.rflags,
             cr0: sregs.cr0,
+            cr2: sregs.cr2,
             cr3: sregs.cr3,
             cr4: sregs.cr4,
-            fs_base: sregs.fs.base,
-        }))
+            sysenter_cs: _msrs.entries[0].data,
+            sysenter_esp: _msrs.entries[1].data,
+            sysenter_eip: _msrs.entries[2].data,
+            msr_efer: _msrs.entries[3].data,
+            msr_star: _msrs.entries[4].data,
+            msr_lstar: _msrs.entries[5].data,
+            efer: sregs.efer,
+            apic_base: sregs.apic_base,
+            cs: segment_reg {
+                base: sregs.cs.base,
+                limit: sregs.cs.limit,
+                selector: sregs.cs.selector,
+            },
+            ds: segment_reg {
+                base: sregs.ds.base,
+                limit: sregs.ds.limit,
+                selector: sregs.ds.selector,
+            },
+            es: segment_reg {
+                base: sregs.es.base,
+                limit: sregs.es.limit,
+                selector: sregs.es.selector,
+            },
+            fs: segment_reg {
+                base: sregs.fs.base,
+                limit: sregs.fs.limit,
+                selector: sregs.fs.selector,
+            },
+            gs: segment_reg {
+                base: sregs.gs.base,
+                limit: sregs.gs.limit,
+                selector: sregs.gs.selector,
+            },
+            ss: segment_reg {
+                base: sregs.ss.base,
+                limit: sregs.ss.limit,
+                selector: sregs.ss.selector,
+            },
+            tr: segment_reg {
+                base: sregs.tr.base,
+                limit: sregs.tr.limit,
+                selector: sregs.tr.selector,
+            },
+            ldt: segment_reg {
+                base: sregs.ldt.base,
+                limit: sregs.ldt.limit,
+                selector: sregs.ldt.selector,
+            },
+
+        })))
+
+            
+    }
+
+    fn write_registers(&self, vcpu: u16, value: u64, reg: u64) -> Result<(), Box<dyn Error>> {
+        let (mut regs, sregs, _msrs) = self.kvmi.get_registers(vcpu)?;
+        match reg {
+            RAX => regs.rax=value,
+            RBX => regs.rbx=value,
+            RCX => regs.rcx=value,
+            RDX => regs.rdx=value,
+            RSI => regs.rsi=value,
+            RDI => regs.rdi=value,
+            RSP => regs.rsp=value,
+            RBP => regs.rbp=value,
+            R8 => regs.r8=value,
+            R9 => regs.r9=value,
+            R10 => regs.r10=value,
+            R11 => regs.r11=value,
+            R12 => regs.r12=value,
+            R13 => regs.r13=value,
+            R14 => regs.r14=value,
+            R15 => regs.r15=value,
+            RIP => regs.rip=value,
+            RFLAGS => regs.rflags=value,
+            _ => println!("wrong choice"),
+            
+        }
+        self.kvmi.set_registers(vcpu,&mut regs)?;
+        Ok(())
+       
     }
 
     fn pause(&mut self) -> Result<(), Box<dyn Error>> {
@@ -139,6 +217,7 @@ impl Introspectable for Kvm {
             InterceptType::Cr(micro_cr_type) => {
                 let kvmi_cr = match micro_cr_type {
                     CrType::Cr0 => KVMiCr::Cr0,
+                    CrType::Cr2 => KVMiCr::Cr2,
                     CrType::Cr3 => KVMiCr::Cr3,
                     CrType::Cr4 => KVMiCr::Cr4,
                 };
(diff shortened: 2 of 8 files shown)


arnabcs17b006:
Do not merge until previous ones have been merged!

Cargo.toml
```diff
@@ -24,7 +24,7 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", optional = true }
+kvmi = { git = "https://github.com/arnabcs17b006/kvmi", optional = true }
```
rageagainsthepc, line 27: Do you have unmerged changes in your own kvmi fork?

c_examples/mem-dump.c
```diff
@@ -47,4 +47,4 @@ int main(int argc, char* argv[]) {
     dump_memory(driver, argv[1]);
     microvmi_destroy(driver);
     return 0;
-}
+}
```
rageagainsthepc, line 50: Please install the editorconfig plugin so we don't see such unnecessary whitespace changes again.

src/driver/kvm.rs
```diff
@@ -261,13 +429,602 @@ mod tests {
...
+    #[test_case(2; "two vcpus")]
+    #[test_case(16; "sixteen vcpus")]
+    fn test_pause_vcpu(vcpu_count: u32) {
+        let mut kvmi_mock = MockKVMi::default();
+        kvmi_mock.expect_init().times(1).returning(|_| Ok(()));
+        kvmi_mock
+            .expect_get_vcpu_count()
+            .times(3)
+            .returning(move || Ok(vcpu_count));
+        for vcpu in 0..vcpu_count {
```
rageagainsthepc, line 503: You should move all expectations that belong to the new and the drop method into a separate method like e.g. `setup_kvmi_mock()`. You are also too specific with your mock setups. Remember, you want to test `pause_vcpu` here. Ideally, a test should not fail for any other reason than something being wrong with the part that is being tested. It shouldn't fail for example if `get_vcpu_count()` has been called 4 instead of 3 times.

src/driver/kvm.rs
```diff
@@ -261,13 +429,602 @@ mod tests {
...
         }
 
         let result = Kvm::new("some_vm", kvmi_mock);
 
         assert!(result.is_ok(), "Expected ok, got error instead!");
     }
 
+    #[test_case(1; "single vcpu")]
+    #[test_case(2; "two vcpus")]
+    #[test_case(16; "sixteen vcpus")]
```
rageagainsthepc, line 495: Don't copy this part. That just over-complicates the tests. I did this in my test to verify that the initialization is correct even for multiple vcpus. You don't need to verify that again and again ;)

src/driver/kvm.rs
```diff
@@ -261,13 +429,602 @@ mod tests {
...
+                    eq(vcpu as u16),
+                    function(|x| matches!(x, KVMiInterceptType::Pagefault)),
+                    eq(false),
+                )
+                .times(1)
+                .returning(|_, _, _| Ok(()));
+        }
+        kvmi_mock.expect_pause().times(1).returning(|| Ok(()));
+        let mut kvm = Kvm::new("some_vm", kvmi_mock).expect("failed to create driver");
+        let result = Kvm::pause(&mut kvm);
```
rageagainsthepc, line 579: Why? What's wrong with the old fashioned `kvm.pause()`

src/driver/kvm.rs
```diff
@@ -261,13 +429,602 @@ mod tests {
...
 
         let result = Kvm::new("some_vm", kvmi_mock);
 
         assert!(result.is_ok(), "Expected ok, got error instead!");
     }
 
+    #[test_case(1; "single vcpu")]
+    #[test_case(2; "two vcpus")]
+    #[test_case(16; "sixteen vcpus")]
+    fn test_pause_vcpu(vcpu_count: u32) {
```
rageagainsthepc, line 496: The naming convention for the tests is `test_feature_under_test()` not `test_method_name()`. You should be more descriptive in your test function names.

src/driver/kvm.rs
```diff
@@ -261,13 +429,602 @@ mod tests {
...
+                .expect_control_events()
+                .with(
+                    eq(vcpu as u16),
+                    function(|x| matches!(x, KVMiInterceptType::Pagefault)),
+                    eq(false),
+                )
+                .times(1)
+                .returning(|_, _, _| Ok(()));
+        }
+        kvmi_mock.expect_pause().times(1).returning(|| Ok(()));
```
rageagainsthepc, line 577: When you write tests, please adhere to the following convention:
```
test_func() {
    setup();

    test();

    assertion();
}
```

Splitting your tests into these three logical blocks improves readability.

Cargo.toml
arnabcs17b006: Yes. I made ffi_event a public member of event struct.

rageagainsthepc: Ah, I see. Could you create a PR for that?

arnabcs17b006: Sure👍

Wenzel:
@arnabcs17b006 same question here ?

arnabcs17b006:
Yes.. It is relevant

Wenzel:
Then I think you need a rebase too

src/driver/kvm.rs
rageagainsthepc: I'm commenting here because I have the feeling that you've forgotten about this.

src/driver/kvm.rs
```diff
@@ -474,6 +455,953 @@ mod tests {
...
+        let mut kvmi_mock = MockKVMi::default();
+        setup_kvmi_mock(&mut kvmi_mock);
+        kvmi_mock.expect_pause().times(0).returning(|| Ok(()));
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        kvm.expect_pause_ev = 1;
```
rageagainsthepc, line 503: I would like to avoid manipulating internal variables. Unit tests should try to verify that the public interface of a class behaves as expected. To this end, private functions are usually not tested explicitely and the internal state of an object should only be modified through external influence. I would rename the test to something like `test_pause_vm_only_once_for_multiple_invocations()` and then test accordingly.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        let timeout = 0;
+        let result = Kvm::listen(&mut kvm, timeout);
+
+        assert!(result.is_ok(), "Expected ok, got error instead!");
```
rageagainsthepc, line 561: In this test it would make sense to verify that listen returns the expected event.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+        let result = Kvm::listen(&mut kvm, timeout);
+
+        assert!(result.is_err(), "Expected error, got ok instead!");
+    }
+
+    #[test]
+    fn test_set_page_access_succeeds() {
+        let mut kvmi_mock = MockKVMi::default();
+        setup_kvmi_mock(&mut kvmi_mock);
+        kvmi_mock
```
rageagainsthepc, line 594: It would also make sense to expect the correct access rights here.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        let paddr: u64 = 0;
+        let result = Kvm::get_page_access(&mut kvm, paddr);
+
+        assert!(result.is_ok(), "Expected ok, got error instead!");
```
rageagainsthepc, line 658: Same here, verify the content.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        let result = Kvm::get_max_physical_addr(&mut kvm);
+
+        assert!(result.is_ok(), "Expected ok, got error instead!");
```
rageagainsthepc, line 701: Verify the content.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+            });
+        kvmi_mock.expect_reply().times(5).returning(|_, _| Ok(()));
+
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        kvm.expect_pause_ev = 5;
```
rageagainsthepc, line 748: try to reach the same effect with manipulating this value directly.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+    #[test]
+    fn test_resume_vcpu_with_multiple_pause_events() {
+        let mut kvmi_mock = MockKVMi::default();
+        setup_kvmi_mock(&mut kvmi_mock);
+        kvmi_mock
+            .expect_wait_and_pop_event()
+            .times(5)
+            .returning(|_| {
+                Ok(Some(KVMiEvent {
+                    vcpu: 0,
```
rageagainsthepc, line 735: Returning 5 pause events for the same vcpu? This is unlikely to happen. You could pass in an outer varable which you increment every time you return something here.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_read_physical_succeeds() {
+        let mut kvmi_mock = MockKVMi::default();
+        setup_kvmi_mock(&mut kvmi_mock);
+        kvmi_mock
+            .expect_read_physical()
+            .times(1)
+            .returning(|_, _| Ok(()));
```
rageagainsthepc, line 849: I would write something into the buffer here and verify that the content at the end.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1009 @@ mod tests {
...
+        )
+        .expect("Failed to create driver");
+        let paddr: u64 = 0;
+        let mut buffer: [u8; 4096] = [0; 4096];
+        let result = Kvm::read_physical(&mut kvm, paddr, &mut buffer);
+        assert!(result.is_err(), "Expected error, got ok instead!");
+    }
+
+    #[test]
+    fn test_write_physical_succeeds() {
```
rageagainsthepc, line 885: Same principle as in read_physical, just the other way around. Expect some specific content in the mocked call.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1014 @@ mod tests {
...
+            .times(0)
+            .returning(|_, _, _| Ok(()));
+        kvmi_mock
+            .expect_control_events()
+            .times(0)
+            .returning(|_, _, _| Ok(()));
+        kvmi_mock
+            .expect_control_events()
+            .times(0)
+            .returning(|_, _, _| Ok(()));
```
rageagainsthepc, line 1058: That shouldn't be necessary. The test would fail anyway if you called these functions without setting them up first. Also, you are asserting to much here. The test shouldn't fail for reasons that don't belong to the hypothesis.

src/driver/kvm.rs
```diff
@@ -474,6 +455,1014 @@ mod tests {
...
+        kvmi_mock
+            .expect_control_events()
+            .with(
+                eq(0),
+                function(|x| matches!(x, KVMiInterceptType::Msr)),
+                eq(false),
+            )
+            .times(1)
+            .returning(|_, _, _| Ok(()));
+        kvmi_mock
```
rageagainsthepc, line 1360: @Wenzel is it logically possible for toggling pagefault interception on to fail if called a second time?

src/driver/kvm.rs
```diff
@@ -474,6 +455,1014 @@ mod tests {
...
+
+        assert!(result.is_ok(), "Expected ok, got error instead!");
+    }
+
+    #[test]
+    fn test_pf_intercept_fails_if_kvmi_conrol_events_fails() {
+        let mut kvmi_mock = MockKVMi::default();
+        kvmi_mock.expect_init().times(1).returning(|_| Ok(()));
+        kvmi_mock.expect_get_vcpu_count().returning(move || Ok(1));
+        kvmi_mock
```
rageagainsthepc, line 1324: why have you replaced setup_kvmi() with this? :scream:

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+
+        assert!(result.is_err(), "Expected error, got ok instead!");
+    }
+
+    #[test]
+    fn test_write_physical_succeeds() {
+        let mut kvmi_mock = MockKVMi::default();
+        setup_kvmi_mock(&mut kvmi_mock);
+        let paddr: u64 = 0;
+        let mut buffer: [u8; 4096] = [0; 4096];
```
rageagainsthepc, line 911: Of all the magical values you could think of for testing, 0 is probably the worst one since it is often used for default initialization.

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+        kvmi_mock
+            .expect_write_physical()
+            .times(1)
+            .with(eq(0), function(|x| matches!(x, _buffer)))
```
rageagainsthepc, line 915: 1. You should compare with `paddr`, not with hardcoded `0`
2. `matches!` is only for enum matching. What you are doing here makes no sense and will always return true. You could write `matches!(x, bla)` for the same effect. I encourage you to look at the expanded macro: `cargo expand --features kvm --lib --tests` (You need to install cargo-expand first)

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+            });
+        let kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+
+        let paddr: u64 = 0;
+        let mut buffer: [u8; 4096] = [0; 4096];
```
rageagainsthepc, line 950: belongs to arrange block

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+            .times(1)
+            .returning(|_, _| Ok(()));
+        let kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+
+        let vcpu: u16 = 0;
```
rageagainsthepc, line 972: same here

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+            ))
+        });
+        let kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+
+        let vcpu: u16 = 0;
```
rageagainsthepc, line 996: same here

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+            .times(1)
+            .return_once(move |_| Ok((regs, sregs, msrs)));
+        let kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
+        let vcpu: u16 = 0;
+        let result = kvm.read_registers(vcpu);
```
rageagainsthepc, line 1020: belongs to act block

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+    }
+
+    #[test]
+    fn test_cr_intercept_fails_if_kvmi_control_cr_fails() {
+        let mut kvmi_mock = MockKVMi::default();
+        kvmi_mock
+            .expect_control_cr()
+            .times(1)
+            .with(eq(0), function(|x| matches!(x, KVMiCr::Cr3)), eq(true))
+            .times(1)
```
rageagainsthepc, line 1079: 2x times()?

src/driver/kvm.rs
```diff
@@ -474,6 +455,884 @@ mod tests {
...
+                    "something went wrong",
+                ))
+            });
+        setup_kvmi_mock(&mut kvmi_mock);
+        let mut kvm = Kvm::new(
+            "some_vm",
+            kvmi_mock,
+            Some(DriverInitParam::KVMiSocket("/tmp/introspector".to_string())),
+        )
+        .expect("Failed to create driver");
```
rageagainsthepc, line 1092: Something that came to my mind: This part is also almost always the same right? Perhaps you could write another function the yields a kvm object completely initialized and set up. If you have the time, I don't deem this necessary. But it would certainly reduce the amount of duplicate setup code alot.

rageagainsthepc: Ah, you know what? Don't bother with that. You have enough on your plate as it is.

PR #85, opened 2020-06-28
simplify c api

rageagainsthepc:
Another box around the box could have spared us a lot of work. ;)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", optional = true }
+kvmi = { git = "https://github.com/Wenzel/kvmi", rev = "b9d8df6730b46f6a00c9ba6e1adcdd5434f613d7", optional = true }
 fdp = { git = "https://github.com/Wenzel/fdp", optional = true }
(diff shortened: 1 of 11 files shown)


src/driver/hyperv.rs
```diff
@@ -255,12 +255,6 @@ impl HyperV {
     }
 }
 
-impl Introspectable for HyperV {
```
Wenzel, line 258: I think you need to leave an empty Impl for Hyper-V, where all methods returns unimplemented(), otherwise the CI fails

rageagainsthepc: Actually, this wasn't intentional. I did not want to delete the whole impl block. :scream:

Cargo.toml
```diff
@@ -24,7 +24,7 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", optional = true }
+kvmi = { git = "https://github.com/Wenzel/kvmi", rev = "b9d8df6730b46f6a00c9ba6e1adcdd5434f613d7", optional = true }
```
Wenzel, line 27: What happens if we use master branch ?
I don't want to commit a specific revision here.
Also, we should switch to published crates.

rageagainsthepc: Libmicrovmi is currently broken on master after the last kvmi merge. Why not pin a commit? It is functionally equivalent to pinning a version and we don't have published crates right now.

rageagainsthepc: I consider this a temporary solution until we either merge a PR that is compatible with the latest changes in kvmi or start publishing crates on crates.io

Wenzel: now it makes sense

Wenzel:
Can I merge this one ?

rageagainsthepc:
Yes, go ahead ;)

PR #87, opened 2020-07-11
msr events

arnabcs17b006:


src/driver/kvm.rs
```diff
@@ -209,6 +212,9 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
                 };
                 Ok(self.kvmi.control_cr(vcpu, kvmi_cr, enabled)?)
             }
+            InterceptType::Msr(micro_msr_type) => {
```
rageagainsthepc, line 215: What does the "micro" stand for in `micro_msr_type`?

src/driver/kvm.rs
```diff
@@ -229,6 +235,11 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
                         new,
                         old,
                     },
+                    KVMiEventType::Msr { msr_type, new, old } => EventType::Msr {
+                        msr_type,
+                        new,
+                        old,
+                    },
                     KVMiEventType::PauseVCPU => panic!("Unexpected PauseVCPU event. It should have been popped by resume VM. (Did you forget to resume your VM ?)"),
                     _ => unimplemented!()
```
rageagainsthepc, line 244: I think I did that temporarily so it compiles. Now that you have added all possible match arms you can just remove this again.

examples/msr-events.rs
```diff
@@ -0,0 +1,111 @@
...
+    drv.resume().expect("Failed to resume VM");
+}
+
+fn main() {
+    env_logger::init();
+
+    let matches = parse_args();
+
+    let domain_name = matches.value_of("vm_name").unwrap();
+    let registers: Vec<_> = matches.values_of("register").unwrap().collect();
```
rageagainsthepc, line 49: I think it's not really a good example to just pass integers as msr indices for two reasons:
1. You have to look those indices up.
2. Currently we are only supporting a very limited set of those indices. The user would have to trial an error to find out which indices actually work.
Therefore I suggest that you match strings like "lstar" to their respective indices. We only support 6 msr values currently so it's not hard to do. Additionally you should write what values are possible into the argument description.

examples/msr-events.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    let mut vec_msr = Vec::new();
+    for reg_str in registers {
+        let without_prefix = reg_str.trim_start_matches("0x");
+        let msr = u32::from_str_radix(without_prefix, 16).unwrap();
+        vec_msr.push(msr);
+    }
```
rageagainsthepc, line 56: nitpick: Put this into a separate method.

examples/msr-events.rs
```diff
@@ -0,0 +1,111 @@
...
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Instant;
+use std::u32;
+
+use microvmi::api::{EventReplyType, EventType, InterceptType, Introspectable};
+
+fn parse_args() -> ArgMatches<'static> {
+    App::new(file!())
+        .version("0.3")
```
rageagainsthepc, line 13: nitpick: Your first version is already 0.3? ;)

examples/msr-events.rs
```diff
@@ -0,0 +1,111 @@
...
+        .author("Mathieu Tarral")
```
rageagainsthepc, line 14: nitpick: Mathieu is not the author of this file, you are. Either remove this or enter your own name.

PR #88, opened 2020-07-11
breakpoint event support added

arnabcs17b006:


examples/interrupt-events.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn parse_args() -> ArgMatches<'static> {
+    App::new(file!())
+        .version("0.1")
+        .about("Watches interrupt VMI events")
+        .arg(Arg::with_name("vm_name").index(1).required(true))
+        .get_matches()
+}
+
+fn get_interrupts(drv: &mut Box<dyn Introspectable>, enabled: bool) {
```
rageagainsthepc, line 19: nitpick: "get" in the function name usually implies that your function yields a return value. You don't return anything here so this function name is misleading.

examples/interrupt-events.rs
```diff
@@ -0,0 +1,93 @@
...
+    let r = running.clone();
+    ctrlc::set_handler(move || {
+        r.store(false, Ordering::SeqCst);
+    })
+    .expect("Error setting Ctrl-C handler");
+
+    println!("Initialize Libmicrovmi");
+    let mut drv: Box<dyn Introspectable> = microvmi::init(domain_name, None);
+
+    //enable int3 interception
```
rageagainsthepc, line 51: As a general rule of thumb: Try to make your function names descriptive enough so you don't need these kinds of comments.

PR #89, opened 2020-07-12
KVM: pagefault event support added

arnabcs17b006:


examples/mem-events.rs
```diff
@@ -0,0 +1,110 @@
...
+
+fn parse_args() -> ArgMatches<'static> {
+    App::new(file!())
+        .version("0.1")
+        .about("Watches memory VMI events")
+        .arg(Arg::with_name("vm_name").index(1).required(true))
+        .get_matches()
+}
+
+fn get_pagefaults(drv: &mut Box<dyn Introspectable>, enabled: bool) {
```
rageagainsthepc, line 22: Please choose a better name for this function.

src/driver/kvm.rs
```diff
@@ -162,6 +165,41 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
...
+
+    fn set_page_access(&self, paddr: u64, access: u8) -> Result<(), Box<dyn Error>> {
+        Ok(self
+            .kvmi
+            .set_page_access(paddr, access)
+            .unwrap()
+            .try_into()?)
+    }
+
+    fn handle_pf_event(&self, paddr: u64, pf_access: u8) -> Result<(), Box<dyn Error>> {
```
rageagainsthepc, line 180: Why have you added this function to the api? This looks to me like it is specifically tailored to your use case. In libmicrovmi we aim to provide a low level interface, not to implement convenience functions to handle events in a specific way.

src/driver/kvm.rs
```diff
@@ -162,6 +165,41 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
         Ok(())
     }
 
+    fn get_page_access(&self, paddr: u64) -> Result<u8, Box<dyn Error>> {
+        Ok(self.kvmi.get_page_access(paddr).unwrap().try_into()?)
+    }
+
+    fn set_page_access(&self, paddr: u64, access: u8) -> Result<(), Box<dyn Error>> {
```
rageagainsthepc, line 172: I would prefer an enum for `access`

examples/mem-events.rs
```diff
@@ -0,0 +1,138 @@
...
+    for vcpu in 0..drv.get_vcpu_count().unwrap() {
+        drv.toggle_intercept(vcpu, intercept, enabled)
+            .expect(&format!("Failed to enable page faults"));
+    }
+
+    drv.resume().expect("Failed to resume VM");
+}
+
+fn handle_pf_event(drv: &mut Box<dyn Introspectable>, gpa: u64, access: u8) {
+    let read_permission = KVMiPageAccess::PageAccessR as u8;
```
rageagainsthepc, line 37: You are breaking the abstraction here. The example should be hypervisor agnostic.

src/api.rs
```diff
@@ -74,6 +74,37 @@ pub enum Registers {
...
+    R,
+    W,
+    X,
+    RW,
+    RX,
+    WX,
+    RWX,
+}
+
+impl From<u8> for Access {
```
rageagainsthepc, line 90: Afais this is KVM specific so it does not belong in this file. Also, if you need to account for error cases you should implement `TryFrom` instead of `From`

src/driver/kvm.rs
rageagainsthepc: You are just passing the access permissions without any abstraction here. This is a bad idea because hypervisors have differing conventions for access permissions. Now, I understand that you want to be able to do arithmetics with the values so a plain enum would problably not be the best fit. Instead I suggest you have a look at the `bitflags` crate.

src/api.rs
```diff
@@ -1,5 +1,66 @@
...
+        const NIL = 0b00000000;
+        const DUMMY = 0b11111111;
+        const RW = Self::R.bits | Self::W.bits;
+        const WX = Self::W.bits | Self::X.bits;
+        const RX = Self::R.bits | Self::X.bits;
+        const RWX = Self::R.bits | Self::W.bits | Self::X.bits;
+    }
+}
+
+impl TryFrom<Access> for u8 {
```
rageagainsthepc, line 18: No need for `TryFrom` here since there is no way this conversion can fail. I would simply implement the `Into` trait for `Access`.

src/api.rs
```diff
@@ -1,5 +1,66 @@
...
+            5 => Access::RX,
+            6 => Access::WX,
+            7 => Access::RWX,
+            _ => Access::DUMMY,
+        };
+        if p_access == Access::DUMMY {
+            Err("not valid access!")
+        } else {
+            Ok(p_access)
+        }
```
rageagainsthepc, line 60: It looks to me like you have introduced the `DUMMY` value for the sole purpose of being able to match on the fallthrough case. This is not good style. Why not simply:
```suggestion
        match access {
            0 => Ok(Access::NIL),
            1 => Ok(Access::R),
            2 => Ok(Access::W),
            3 => Ok(Access::RW),
            4 => Ok(Access::X),
            5 => Ok(Access::RX),
            6 => Ok(Access::WX),
            7 => Ok(Access::RWX),
            _ => Err("invalid access value"),
        }
```

rageagainsthepc: Also you should not use plain integers if there are `KVMI_PAGE_ACCESS` definitions available in the kvmi crate.

examples/mem-events.rs
```diff
@@ -0,0 +1,131 @@
...
+    println!("{} memory events", status_str);
+    for vcpu in 0..drv.get_vcpu_count().unwrap() {
+        drv.toggle_intercept(vcpu, intercept, enabled)
+            .expect(&format!("Failed to enable page faults"));
+    }
+
+    drv.resume().expect("Failed to resume VM");
+}
+
+fn handle_pf_event(drv: &mut Box<dyn Introspectable>, gpa: u64, access: Access) {
```
rageagainsthepc, line 36: This function is way to complicated for what it actually does. Correct me if I am wrong but you want allow the access that caused the pagefault here, right? Why not call `get_page_access`, add the permission bit that caused the page fault, call `set_page_access` and be done with it?

examples/mem-events.rs
```diff
@@ -0,0 +1,131 @@
+use clap::{App, Arg, ArgMatches};
+use colored::*;
+use env_logger;
+use std::convert::TryInto;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::Arc;
+use std::time::Instant;
+
+//use microvmi::api::{Access, EventReplyType, EventType, InterceptType, Introspectable, };
```
rageagainsthepc, line 9: You probably forgot to revert this?

examples/mem-events.rs
```diff
@@ -0,0 +1,110 @@
...
+                let pagefault_output = format!("pagefault occurred!").color("blue");
+                let pf_access: Access = access.try_into().unwrap();
+                println!(
+                    "[{}] {} - {}:   gva = 0x{:x}    gpa = 0x{:x}    access = {:?} ",
+                    ev_nb_output, vcpu_output, pagefault_output, gva, gpa, pf_access
+                );
+                let mut p_access = drv.get_page_access(gpa).expect("Failed to get page access");
+                p_access |= pf_access;
+                drv.set_page_access(gpa, p_access)
+                    .expect("Failed to set page access");
```
rageagainsthepc, line 92: nitpick: Using a function or adding a comment here would be helpful so it is immediately clear what you intend to do.

src/driver/kvm.rs
```diff
@@ -1,17 +1,54 @@
...
 use std::vec::Vec;
 
 use crate::api::{
-    CrType, Event, EventReplyType, EventType, InterceptType, Introspectable, Registers, SegmentReg,
-    X86Registers, PAGE_SHIFT,
+    Access, CrType, Event, EventReplyType, EventType, InterceptType, Introspectable, Registers,
+    SegmentReg, X86Registers, PAGE_SHIFT,
 };
 
+impl TryFrom<Access> for u8 {
```
rageagainsthepc, line 16:
```suggestion
impl Into<u8> for Access {
```

rageagainsthepc: Again, we should really use `KVMiPageAccess` instead of `u8` everywhere possible so the conversion will be unambiguous.

src/driver/kvm.rs
```diff
@@ -1,17 +1,54 @@
...
+            _ => Err("invalid access value"),
+        }
+    }
+}
+
+impl TryFrom<u8> for Access {
+    type Error = &'static str;
+    fn try_from(access: u8) -> Result<Self, Self::Error> {
+        match access {
+            0 => Ok(Access::NIL),
```
rageagainsthepc, line 39: You should be consistent and use `KVMiPageAccess` here as well.

src/driver/kvm.rs
```diff
@@ -162,6 +199,17 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
...
     }
 
+    fn get_page_access(&self, paddr: u64) -> Result<Access, Box<dyn Error>> {
+        let access = self.kvmi.get_page_access(paddr).unwrap();
+        Ok(access.try_into().unwrap())
+    }
+
+    fn set_page_access(&self, paddr: u64, access: Access) -> Result<(), Box<dyn Error>> {
+        let p_access = access.try_into().unwrap();
+        self.kvmi.set_page_access(paddr, p_access)?;
```
rageagainsthepc, line 209:
```suggestion
        self.kvmi.set_page_access(paddr, access.try_into().unwrap())?;
```

rageagainsthepc:
fyi: This build fails right now because there is a bug in actions-rs/install https://github.com/actions-rs/install/issues/6

examples/mem-events.rs
```diff
@@ -0,0 +1,109 @@
...
+                    _ => panic!("Not pf event"),
+                };
+                let ev_nb_output = format!("{}", i).cyan();
+                let vcpu_output = format!("VCPU {}", ev.vcpu).yellow();
+                let pagefault_output = format!("pagefault occurred!").color("blue");
+                println!(
+                    "[{}] {} - {}:   gva = 0x{:x}    gpa = 0x{:x}    access = {:?} ",
+                    ev_nb_output, vcpu_output, pagefault_output, gva, gpa, access
+                );
+                let mut p_access = drv.get_page_access(gpa).expect("Failed to get page access");
```
rageagainsthepc, line 87: nitpick: The naming of `p_access` and `access` is confusing imho.

rageagainsthepc:
lgtm @Wenzel ?

Wenzel:
What's happening with the security audit ?
it ran for 5h and failed ??! :thinking:

arnabcs17b006:
To be precise, it got aborted.

rageagainsthepc:
I think we shouldn't wait any longer for the bug to get fixed. I think the important thing is that the build jobs are passing.

Wenzel:
Does this actually work for you @arnabcs17b006 ?
It freezes my VM when it starts modifying the memory access on the whole memory.
I couldn't event send a CTRL-C, weird.
Capture d’écran de 2020-08-02 12-00-39

Wenzel:
@arnabcs17b006 this PR needs some rebase on master now that the documentation PR has been merged.

src/api.rs
```diff
@@ -192,10 +205,30 @@ pub trait Introspectable {
     /// # Arguments
     /// * 'vcpu' - vcpu id for which the value of registers are to be dumped as the argument
     ///
+    ///
```
rageagainsthepc, line 208: Is this change here intentional?

Wenzel:
I confirm this code works for me:
Capture d’écran de 2020-08-19 12-50-01

I need a final review and we can merge

PR #92, opened 2020-08-02
examples: msr-events intercepts 6 MSRs by default

Wenzel:
rework msr-events example

* intercepts 6 MSRs by default
* can specify custom set of MSRs

### DIFF
--- a/examples/msr-events.rs
+++ b/examples/msr-events.rs
@@ -8,17 +8,34 @@ use std::u32;
 
 use microvmi::api::{EventReplyType, EventType, InterceptType, Introspectable};
 
+// default set of MSRs to be intercepted
+const DEFAULT_MSR: [u32; 6] = [0x174, 0x175, 0x176, 0xc0000080, 0xc0000081, 0xc0000082];
+
 fn parse_args() -> ArgMatches<'static> {
     App::new(file!())
-        .version("0.1")
-        .about("Watches msr register VMI events")
-        .arg(Arg::with_name("vm_name").index(1).required(true))
+        .version("0.2")
+        .about(
+            "Watches MSR register VMI events\n\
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", optional = true }
+kvmi = { git = "https://github.com/Wenzel/kvmi", rev = "dd10135a27bb3658d399dfb5477299ca0f4baeac", optional = true }
```
rageagainsthepc, line 27: Why do you pin this commit?

rageagainsthepc:
So you want to pass raw integers instead of register names? Why? Imho thats not handy at all, especially not for an example.

Wenzel:
Because it doesn't limit this example to the 6 hardcoded registers we have defined, which are now intercepted by default.

The end goal is to use the x86 crate with their msr definitions, so you can pass (any) string again.

I need to open an issue to ask them to split it into a separate crate, bevause it requires nightly rust now.

rageagainsthepc:
But those are all the registers we are able to intercept right now. If we want to intercept more, we would need to change the code of kvmi as well.

rageagainsthepc:
My bad, I think I mixed something up. It's perfectly possible to intercept arbitrary MSRs. Sorry about the confusion.

Cargo.toml
Wenzel: because using `master` broke the CI

PR #93, opened 2020-08-06
Xen: event support

arnabcs17b006:


Cargo.toml
```diff
@@ -20,20 +20,23 @@ hyper-v = ["winapi", "widestring", "ntapi", "vid-sys"]
...
 log = "0.4.8"
-env_logger = "0.7.1"
 libc = { version = "0.2.58", optional = true }
-xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
+xenctrl = { git = "https://github.com/arnabcs17b006/xenctrl", branch = "write_register", optional = true }
 xenstore = { git = "https://github.com/Wenzel/xenstore", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
-kvmi = { git = "https://github.com/Wenzel/kvmi", rev = "dd10135a27bb3658d399dfb5477299ca0f4baeac", optional = true }
+xenevtchn = { git = "https://github.com/arnabcs17b006/xenevtchn", branch = "event-notification"}
+xenvmevent-sys = { git = "https://github.com/Wenzel/xenvmevent-sys"}
```
rageagainsthepc, line 28: Why is there a sys crate in here? We should probably create a safe API layer for it. @Wenzel ?

rageagainsthepc, line 28: You should mark those dependencies as optional, otherwise the CI will always fail.

Wenzel: Yes, we should have
but it's a bit late now.
Let'd keep it like this and improve in the future.

examples/cr-events.rs
```diff
@@ -101,7 +102,7 @@ fn main() {
     // listen
     let mut i: u64 = 0;
     while running.load(Ordering::SeqCst) {
-        let event = drv.listen(1000).expect("Failed to listen for events");
+        let event = drv.listen(10).expect("Failed to listen for events");
```
Wenzel, line 105: why did you changed the timeout ?

examples/cr-events.rs
```diff
@@ -42,10 +42,11 @@ fn toggle_cr_intercepts(drv: &mut Box<dyn Introspectable>, vec_cr: &Vec<CrType>,
         let intercept = InterceptType::Cr(*cr);
         let status_str = if enabled { "Enabling" } else { "Disabling" };
         println!("{} intercept on {:?}", status_str, cr);
-        for vcpu in 0..drv.get_vcpu_count().unwrap() {
-            drv.toggle_intercept(vcpu, intercept, enabled)
-                .expect(&format!("Failed to enable {:?}", cr));
-        }
+        //for vcpu in 0..drv.get_vcpu_count().unwrap() {
```
Wenzel, line 45: why did you commented the loop ?

src/driver/xen.rs
```diff
@@ -164,6 +184,86 @@ impl Introspectable for Xen {
         }))
     }
 
+    fn listen(&mut self, timeout: u32) -> Result<Option<Event>, Box<dyn Error>> {
+        let fd = self.xev.xenevtchn_fd()?;
```
Wenzel, line 188: you don't need to recreate the fd and the poll structure everytime.
create it once, store it, and use it here.

src/driver/xen.rs
```diff
@@ -164,6 +184,86 @@ impl Introspectable for Xen {
...
+        let fd_struct = PollFd::new(fd, PollFlags::POLLIN | PollFlags::POLLERR);
+        let mut fds = [fd_struct];
+        let mut vcpu: u16 = 0;
+        let mut event_type = unsafe { mem::MaybeUninit::<EventType>::zeroed().assume_init() };
+        let poll_result = poll(&mut fds, timeout.try_into().unwrap()).unwrap();
```
Wenzel, line 193: forward the error, either with an explicit match, or the `?` (try) operator

Wenzel, line 193: use a `match` on the poll result to handle each possible value, and `_` for all the others.

src/driver/xen.rs
```diff
@@ -164,6 +184,86 @@ impl Introspectable for Xen {
...
+        let mut pending_event_port = -1;
+        if poll_result == 1 {
+            pending_event_port = self.xev.xenevtchn_pending()?;
+            if pending_event_port != -1 {
+                self.xev
+                    .xenevtchn_unmask(pending_event_port.try_into().unwrap())?;
```
Wenzel, line 199: - forward the error.
- use a match statement to handle all possible values
- is pending_event_port is -1, it is a failure, and your code continues its execution. it shouldn't

src/driver/xen.rs
```diff
@@ -164,6 +184,86 @@ impl Introspectable for Xen {
...
+            }
+        }
+        let back_ring_ptr = &mut self.back_ring;
+        let mut flag = false;
```
Wenzel, line 203: `flag` doesn't mean anything.

src/driver/xen.rs
```diff
@@ -164,6 +184,86 @@ impl Introspectable for Xen {
...
+        if poll_result > 0
+            && self.xev.get_bind_port() == pending_event_port
+            && RING_HAS_UNCONSUMED_REQUESTS!(back_ring_ptr) != 0
+        {
+            flag = true;
+            let req = self.xc.get_request(back_ring_ptr)?;
+            if req.version != VM_EVENT_INTERFACE_VERSION {
+                panic!("version mismatch");
```
Wenzel, line 211: don't panic, it's a library, you should return an Error

src/driver/xen.rs
Wenzel: actually, you don't need it. your `if` statement can return a value, which is your Event, or None.
just call notify, and then Ok on the value returned by the if

Wenzel (changes requested):

PR #94, opened 2020-08-09
actions: remove tool cache from security audit

Wenzel:


Wenzel:
@rageagainsthepc you can also see here that without the second commit pinning the kvmi crate, master doesn't compile :)

PR #97, opened 2020-08-11
actions: do not fail CI upon codecov error

Wenzel:
@rageagainsthepc will this fix our CI ?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -113,7 +113,7 @@ jobs:
         with:
           file: ${{ steps.coverage.outputs.report }}
           flags: unittests
-          fail_ci_if_error: true
+          fail_ci_if_error: false
 
   virtualbox:
     runs-on: ubuntu-latest


Wenzel:
I just found this flag by chance while reviewing the CI configuration.
Maybe it works

rageagainsthepc:
Huh, I knew about this flag, but I was convinced that it only refers to the github action where it is set. Still, I would like to know why codecov has errors in the first place. Oh well... :shrug:

Wenzel:
So, it didn't work :crying_cat_face: 
https://github.com/Wenzel/libmicrovmi/pull/70

Wenzel:
Looking at https://github.com/codecov/codecov-action:

_Specify if CI pipeline should fail when Codecov runs into errors during upload._

So wrong flag :)

Wenzel:
See https://github.com/Wenzel/libmicrovmi/pull/99

rageagainsthepc:
> Looking at https://github.com/codecov/codecov-action:
> 
> _Specify if CI pipeline should fail when Codecov runs into errors during upload._
> 
> So wrong flag :)

So my initial understanding of this flag was correct after all. :)

PR #101, opened 2020-08-16
actions: remove codecov as it breaks the CI

Wenzel:
Proposal: remove codecov as it breaks our CI and we don't know how to fix it.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,15 +105,6 @@ jobs:
           crate: grcov
           version: latest
           use-tool-cache: true
-      - name: generate coverage report
-        id: coverage
-        uses: actions-rs/grcov@v0.1
-      - name: upload coverage report to codecov.io
-        uses: codecov/codecov-action@v1
-        with:
-          file: ${{ steps.coverage.outputs.report }}
-          flags: unittests
-          fail_ci_if_error: false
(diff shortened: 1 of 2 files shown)


codecov.yaml
```diff
@@ -1,7 +0,0 @@
-coverage:
-  status:
-    patch:
-      default:
-        target: auto
```
rageagainsthepc, line 5: According to the documentation this might help:
`target: 0%`

rageagainsthepc: It's the baseline of coverage that the commit must meet (Usually derived from master).

.github/workflows/build.yml
```diff
@@ -105,15 +105,6 @@ jobs:
           crate: grcov
           version: latest
           use-tool-cache: true
-      - name: generate coverage report
```
rageagainsthepc, line 108: If you are not uploading the coverage anymore there is no point in generating it in the first place.

codecov.yaml
rageagainsthepc: Or even better: you can set a flag `ìnformational: true` This should make the CI pass no matter what.

Wenzel:
already fixed by https://github.com/Wenzel/libmicrovmi/pull/102

PR #104, opened 2020-08-16
actions: fix duplicated builds

### LINKED ISSUE #103: Github Actions: avoid dual build problem
At the moment we are triggering a dual build for internal pull requests (a PR originating from this repo, to master branch for example).

After a quick search, it's possible to avoid this. examples:
- https://github.com/decidim/decidim/pull/5883
- https://github.com/microsoft/vscode/pull/83158

I recommend to limit the `push` to `master` branch and let `pull_request` trigger on any branches.

Wenzel:
fix #103

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,12 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - "*"
 
 jobs:
   dummy:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - "*"
 
 jobs:
   dummy:
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -1,6 +1,12 @@
 name: Doc
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - "*"
 
 jobs:
   doc:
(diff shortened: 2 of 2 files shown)


rageagainsthepc:
So what happens when a branch is pushed after a PR has been created? Afaik this is not automatically retested, right?

.github/workflows/build.yml
```diff
@@ -1,6 +1,12 @@
 name: Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - "*"
```
rageagainsthepc, line 9: nitpick: You don't need this afaik.

Wenzel:
> So what happens when a branch is pushed after a PR has been created?

As far as I understand, since the PR is opened, any push on the branch will trigger the `pull_request` event.
The goal is to avoid the `push` event to be triggered also.
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request

.github/workflows/build.yml
Wenzel: so an empty key `pull_request` is avalid you think ?
I can try...

Wenzel:
@rageagainsthepc also, see just now,
we have pushed a new commit and the pull_request event has been triggered, even though the PR was already opened.

.github/workflows/build.yml
Wenzel: uh, you were right

Wenzel:
@rageagainsthepc  ?

rageagainsthepc:
I could swear that at some point feature branches were not retested in PRs if you only had the on pull request action. Glad to see that there is a solution that avoids building twice though. ;) This should also fix clippy annotations being posted twice, right?

PR #106, opened 2020-08-16
Capi return bool

### LINKED ISSUE #105: CAPI: return boolean instead of custom Success/Failure values
Libvmi has been built with (too) strong compatibility in mind, and used a `VMI_SUCCESS/FAILURE` as a return type when it could have included `stdbool.h` and returned a boolean.

`stdbool.h` only exists since `C99`:
https://stackoverflow.com/a/1608350/3017219

I think we should not try to go lower than this standard.

Wenzel:
fix #105


codecov.yaml
```diff
@@ -3,3 +3,6 @@ coverage:
     project:
       default:
         informational: true
+    patch:
```
rageagainsthepc, line 6: Why is this needed?

c_examples/pause.c
```diff
@@ -1,17 +1,19 @@
+#define _DEFAULT_SOURCE
```
rageagainsthepc, line 1: What does this do?

codecov.yaml
Wenzel: Fix the codecov CI for `codecov/patch` to not fail the CI

c_examples/pause.c
Wenzel: removes a warning os usleep definition:
https://stackoverflow.com/a/10053817/3017219

Issue #108, opened 2020-08-16
First publication on crates.io


Wenzel:
# This issue tracks what needs to be done for a first publication on crates.io

- [ ] update Cargo.toml
- [ ] convert all git dependencies to crates.io equivalent

## upload git dependencies to crates.io

- [x] xenctrl
  - [x] xenctrl-sys
- [x] xenstore
  - [x] xenstore-sys
- [x] xenforeignmemory
  - [x] xenforeignmemory-sys
- [x] kvmi
  - [x] kvmi-sys
- [x] fdp
  - [x] fdp-sys

Wenzel:
`kvmi-sys` has been officially published: https://crates.io/crates/kvmi-sys :tada:

Wenzel:
`kvmi` has been published: https://crates.io/crates/kvmi

Wenzel:
- fdp-sys has been published: https://crates.io/crates/fdp-sys
- fdp has been published: https://crates.io/crates/fdp

Wenzel:
- xenforeignmemory-sys: https://crates.io/crates/xenforeignmemory-sys
- xenforeignmemory: https://crates.io/crates/xenforeignmemory

Wenzel:
- xenstore-sys: https://crates.io/crates/xenstore-sys
- xenstore: https://crates.io/crates/xenstore

Wenzel:
xenctrl-sys: https://crates.io/crates/xenctrl-sys

Wenzel:
looking at https://github.com/sunng87/cargo-release to automate publication
We can have a look at this PR: https://github.com/ballista-compute/sqlparser-rs/pull/238

Wenzel:
xenctrl: https://crates.io/crates/xenctrl

Wenzel:
Regarding xenstore:
https://github.com/Wenzel/xenstore/pull/6

Wenzel:
xenstore-rs has been published: https://crates.io/crates/xenstore-rs

PR #109, opened 2020-08-16
cargo: fill required fields for publishing

Wenzel:
Fill essential keys required for a publication on crates.io

help fixing #108


Cargo.toml
```diff
@@ -3,6 +3,12 @@ name = "microvmi"
 version = "0.1.0"
 authors = ["Mathieu Tarral <mathieu.tarral@protonmail.com>"]
 edition = "2018"
+license = "GPL-3.0-only"
+description = "A cross-platform, unified, low-level VMI API on top of hypervisors"
```
Wenzel, line 7: @rageagainsthepc I'd like your opinion on the description.
how do you feel when reading it ?
can we do better ?
my english is not incredible.

rageagainsthepc:
```suggestion
description = "A cross-platform, unified, low-level VMI API supporting multiple hypervisors"
```
Perhaps?

rageagainsthepc: Hm, maybe you can scratch that idea. The word "unified" already gives away that multiple hypervisor are supported. lgtm

Wenzel: Unified means to me "abstraction layer", so I'll keep it for now

PR #112, opened 2020-08-17
System Table registers (IDTR/GDTR)

### LINKED ISSUE #111: Registers: add support for IDT register
While trying to integrate libmicrovmi into LibVMI, I noticed that one of the [`KPCR` search algorithm](https://github.com/libvmi/libvmi/blob/704d6a3150c91664c35bd0e6be344a1386856d10/libvmi/os/windows/core.c#L794) required the [`IDTR_BASE`](https://github.com/libvmi/libvmi/blob/master/libvmi/libvmi.h#L326) LibVMI register.

Even though there are fallback algorithms, they failed for unclear reasons.
So I decided to hardcode my `IDTR_BASE` value:
Capture d’écran de 2020-08-18 00-37-36

And `vmi-process-list` example works, on top of `libmicrovmi` :tada: 
Capture d’écran de 2020-08-18 00-38-29

cc @rageagainsthepc 
the current implementation (without `IDTR_BASE` hardcoded) is here:
https://github.com/Wenzel/libvmi/tree/libmicrovmi

Wenzel:
help fix #111 in Xen

- add definition of IDTR/GDTR (https://xem.github.io/minix86/manual/intel-x86-and-64-manual-vol3/o_fe12b1e2a880e0ce-74.html)
- fetch them in Xen
- fetch them in KVM

regs-dump example (with a little tweak to force get_vcpu_count to 1)
Capture d’écran de 2020-08-18 01-16-31

### DIFF
--- a/src/driver/kvm.rs
+++ b/src/driver/kvm.rs
@@ -11,7 +11,7 @@ use std::vec::Vec;
 
 use crate::api::{
     Access, CrType, DriverInitParam, Event, EventReplyType, EventType, InterceptType,
-    Introspectable, Registers, SegmentReg, X86Registers, PAGE_SHIFT,
+    Introspectable, Registers, SegmentReg, SystemTableReg, X86Registers, PAGE_SHIFT,
 };
 
 impl TryFrom<Access> for KVMiPageAccess {
@@ -192,6 +192,14 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
             ss: sregs.ss.into(),
             tr: sregs.tr.into(),
             ldt: sregs.ldt.into(),
+            idt: SystemTableReg {
+                base: sregs.idt.base,
+                limit: sregs.idt.limit,
+            },
+            gdt: SystemTableReg {
+                base: sregs.gdt.base,
+                limit: sregs.idt.limit,
+            },
         }))
     }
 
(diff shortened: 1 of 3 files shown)


src/driver/kvm.rs
```diff
@@ -155,6 +155,14 @@ impl<T: KVMIntrospectable> Introspectable for Kvm<T> {
             ss: sregs.ss.into(),
             tr: sregs.tr.into(),
             ldt: sregs.ldt.into(),
+            idt: SystemTableReg {
```
rageagainsthepc, line 158: nitpick: It would be great if you could implement the `From` trait here similar to the segment regs.

src/driver/xen.rs
```diff
@@ -161,6 +163,14 @@ impl Introspectable for Xen {
                 limit: 0,
                 selector: 0,
             },
+            idt: SystemTableReg {
```
rageagainsthepc, line 166: Same here.

src/driver/kvm.rs
Wenzel: Good idea, thanks !

src/driver/xen.rs
Wenzel: I don't think this one is possible, since you don't know which table you are building in the From impl (idtr/gdtr)
https://xenbits.xen.org/people/ianc/docs/structhvm__hw__cpu.html

rageagainsthepc: Yeah, you're right. I totally forgot that we have one large struct here with all the information regarding registers.

PR #119, opened 2020-08-19
kvm: disable interception of breakpoint events by default

### LINKED ISSUE #118: KVM: disable interception of breakpoint events by default
While reviewing the code:
https://github.com/Wenzel/libmicrovmi/blob/master/src/driver/kvm.rs#L122

We shouldn't be intercepting INT3 by default, unless we are explicitely asked to do so.
if a debugger is running, the INT3 will be catched and put the in the queue, (and if the user never listens, never processed).

Libvmi implementation:
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/kvm/kvm_events.c#L615

Wenzel:
fix #118


rageagainsthepc:
Yeah, makes sense. :+1:

rageagainsthepc (approved):

Issue #125, opened 2020-12-10
initial setup with example code results in a segmentation fault / xen on linux


martinloos:
## Context

As discussed on twitter (@deiderajr) i receive a segmentation fault after using the mem-dump example code (same for other examples).

## Process

This is our testing environment:

[xlinfo.txt](https://github.com/Wenzel/libmicrovmi/files/5676035/xlinfo.txt)

Steps to reproduce the error:

1. Downloaded the repository from github (master branch)
2. cd to libmicrovmi/target/debug/examples
3. RUST_LOG=debug cargo run --features xen --example mem-dump <VM-Name>

Retrieving data with libvmi works (see terminal_libvmi.png).

![terminal_libvmi](https://user-images.githubusercontent.com/51302052/101840355-3a7cbb80-3b44-11eb-8fe4-72db09c40231.png)

## Expected result

Getting a correct .dump file

## Current result

Segmentation fault (see terminal_libmicrovmi.png)

![terminal_libmicrovmi](https://user-images.githubusercontent.com/51302052/101840336-32bd1700-3b44-11eb-8a3f-3dd9eb24b53d.png)

## End

If i missed important information please let me know. Thanks!

Wenzel:
Hi @martinloos , thanks for your detailed report.

I just tested libmirovmi's `mem-dump` example on a Xen VM, and it works on my side:
![Capture d’écran de 2020-12-11 01-01-14](https://user-images.githubusercontent.com/964610/101844379-8af81700-3b4c-11eb-95cd-f67ff85ce4aa.png)

I'm running Xen 4.11.4-pre, on ubuntu 20.04 (a Vagrant image)
libmicrovmi is on latest master, at `6a28dfe8ceb996cb2255ee014f0a68e0677f1223`

I am quite surprised to see a segfault
Can you try to run it with `rust-gdb` and bet a stacktrace ?
https://bitshifter.github.io/rr+rust/index.html#1

martinloos:
I tried running it with rust-gdb --args ./mem-dump <VM-Name> and the following was the output:

![terminal_rust-gdb_output](https://user-images.githubusercontent.com/51302052/101847179-ec22e900-3b52-11eb-8cc7-56e1434ac108.png)

Wenzel:
You need to run the `bt` command to get the stacktrace

martinloos:
[...]

PR #90, opened 2020-07-17
Xen: reading and writing registers

arnabcs17b006:

PR #128, opened 2020-12-14
xen: handle permission denied when reading xenstore entries

### LINKED ISSUE #127: Xen: Handle permission denied error while reading xenstore entries
According to https://github.com/Wenzel/libmicrovmi/issues/125#issuecomment-744378908

When reading xenstore from a dedicated virtual machine, the access is more limited, and reading some entries may fail with a permission denied error.

In Libvmi, NULL entries are ignored.

We should implement the same behavior, handling explicitely the permission denied error, and continue the loop.

Wenzel:
This should help fix #127 and #125 

Depends on this sub-PR:
https://github.com/Wenzel/xenstore/pull/7

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ crate-type = ["cdylib", "lib"]
 
 [features]
 # Xen driver
-xen = ["xenctrl", "xenstore", "xenforeignmemory", "libc"]
+xen = ["xenctrl", "xenstore-rs", "xenforeignmemory", "libc"]
 # KVM driver
 kvm = ["kvmi"]
 # VirtualBox driver
@@ -29,7 +29,7 @@ log = "0.4.8"
 env_logger = "0.7.1"
 libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
-xenstore = { git = "https://github.com/Wenzel/xenstore", rev = 'e8627acfe73186eb7cc34385d531058d3c55643d', optional = true }
+xenstore-rs = { version = "0.3.0", optional = true }
 xenforeignmemory = { git = "https://github.com/Wenzel/xenforeignmemory", optional = true }
 kvmi = { version = "0.2.1", optional = true }
 fdp = { version = "0.1.0", optional = true }
--- a/src/driver/xen.rs
+++ b/src/driver/xen.rs
@@ -3,6 +3,7 @@ use crate::api::{
 };
 use libc::{PROT_READ, PROT_WRITE};
 use std::error::Error;
+use std::io::ErrorKind;
 use xenctrl::consts::{PAGE_SHIFT, PAGE_SIZE};
 use xenctrl::XenControl;
 use xenstore_rs::{XBTransaction, Xs, XsOpenFlags};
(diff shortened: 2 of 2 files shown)


Wenzel:
@martinloos, @dl9rdz, I believe this PR should you fix your permission denied issue.

rageagainsthepc:
@Wenzel lgtm

Cargo.toml
```diff
@@ -29,7 +29,7 @@ log = "0.4.8"
 env_logger = "0.7.1"
 libc = { version = "0.2.58", optional = true }
 xenctrl = { git = "https://github.com/Wenzel/xenctrl", optional = true }
-xenstore = { git = "https://github.com/Wenzel/xenstore", rev = 'e8627acfe73186eb7cc34385d531058d3c55643d', optional = true }
+xenstore = { git = "https://github.com/Wenzel/xenstore", rev = '728ed1366723ae2b1415e6d25aad35883fb75d55', optional = true }
```
rageagainsthepc, line 32: Shouldn't you use your newly published `xenstore-rs` crate here?

Wenzel: I know !
I started my PR by switching to xenstore-rs and it would fail to build in the CI.

Could be a bug with cargo and git dependencies.

I will anyway switch when using the crate published on crates.io

rageagainsthepc: > I will anyway switch when using the crate published on crates.io

That's what I meant. Also, are we doing a clean checkout? Because if there is a cache somewhere, that would explain why the switch to `xenstore-rs` doesn't work.

martinloos:
@Wenzel I can confirm that it fixed the permission denied issue and i was able to get the mem-dump. 

But i couldn't inspect it with Volatility as i get: 

`Volatility was unable to read a requested page:
Page error 0xffff81e13858 in layer primary (Page Fault at entry 0x0 in table page directory)` 

Is there another way to verify the .dump file received?

And another question: Is there a working example for retrieving the process list available? (similar to the one provided by libvmi)

Thank you for the quick fix!

Wenzel:
@martinloos thanks for the feedback

> Volatility was unable to read a requested page: Page error 0xffff81e13858 in layer primary (Page Fault at entry 0x0 in table page directory)

Strange, are you using Volatility3 ?
Maybe open an issue on their repo, they will get back to you.

> And another question: Is there a working example for retrieving the process list available? (similar to the one provided by libvmi)

No, there isn't, since it's a non-goal of `libmicrovmi`.
Retrieving the process list implies that we have some semantic understanding of the operating system.
libmicrovmi solely aims to be an abstraction layer on the VMI APIs.
Building a VMI semantic library can be done on top of libmicrovmi of course.
We intent to stay small and easy to integrate with other applications ;)

martinloos:
@Wenzel Yes we use Volatility3. I'm going to open an issue there. 

Thanks for explaining!

Two last questions (at least for today): 

1) Is the regs-dump example working on your machine? Mine panics at 'not implemented': api.rs L183 called from regs-dump L36. 

I'm afk atm and need to look at this problem once i'm back, but wanted to make sure that the example is working in general.

2) Regarding the cr-events example. Again, I would like to know whether the example works in principle (particularly for CR3 events).

Wenzel:
> Is the regs-dump example working on your machine? Mine panics at 'not implemented': api.rs L183 called from regs-dump L36.

Correct.
We mentored a student with the GSoC program this summer, and adding read / write register support was part of his assignments.
https://medium.com/@cs17b006/gsoc-with-libmicrovmi-4fcecb7fba5d


His PRs have unfortunately been merged yet, lacking personal time to do it
:arrow_right: you can find the regs read/write PR here:
https://github.com/Wenzel/libmicrovmi/pull/90

> Regarding the cr-events example. Again, I would like to know whether the example works in principle (particularly for CR3 events).

It works as long as the driver is implementing the libmicrovmi API

Regarding Xen, again, the events support is still in multiple PRs:
Capture d’écran de 2020-12-15 21-48-19

cheers !

PR #146, opened 2020-12-17
ci: create github release

Wenzel:
fix https://github.com/Wenzel/libmicrovmi/issues/145

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -228,15 +228,58 @@ jobs:
           cd c_examples
           make
 
-  publish:
+  release:
+    # create a Github release
+    # only when
+    # - push on master
+    # - tag starts with 'v*'
     needs: [format, dummy, xen, kvm, virtualbox, hyper-v, c_api]
+    runs-on: ubuntu-20.04
+    # output these value to be used by other jobs so they can add assets
+    outputs:
+      upload_url: ${{ steps.step_upload_url.outputs.upload_url }}
+      version: ${{ steps.get_version.outputs.version }}
+
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+    steps:
+      - uses: actions/checkout@v1
+
+      - name: Get the version
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: install cargo changelog
+        run: cargo install changelog
+
+      - name: generate changelog
+        run: changelog -o changelog.md
+
+      - name: Create a Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.get_version.outputs.version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -228,15 +228,58 @@ jobs:
...
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          tag_name: ${{ steps.get_version.outputs.version }}
+          release_name: ${{ steps.get_version.outputs.version }}
+          body_path: changelog.md
+
+      - id: step_upload_url
+        run: echo "::set-output name=upload_url::${{ steps.create_release.outputs.upload_url }}"
```
rageagainsthepc, line 274: This step us unnecessary. Why not simply use `steps.create_release.outputs.upload_url` directly?

.github/workflows/ci.yml
```diff
@@ -228,15 +228,58 @@ jobs:
...
+        id: get_version
+        run: echo ::set-output name=VERSION::${GITHUB_REF/refs\/tags\//}
+
+      - name: install stable toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+
+      - name: install cargo changelog
```
rageagainsthepc, line 257: This is step could have been joined with the step above.

rageagainsthepc:
I'm wondering, what happens to the version we provide in `Cargo.toml`? :thinking: Is it unused now?

.github/workflows/ci.yml
Wenzel: so any subcommand is a valid toolchain component ?
I wanted to do that at the beginning, then I went to check what a toolchain component was, and ended up on this
https://rust-lang.github.io/rustup/concepts/components.html

As it was not listed, I figured it did not qualify as a component.

Also, I just tried:
~~~
$ rustup component add changelog
error: toolchain 'stable-x86_64-unknown-linux-gnu' does not contain component 'changelog' for target 'x86_64-unknown-linux-gnu'
~~~

.github/workflows/ci.yml
Wenzel: The goal is to transfer the output_url from this job to other jobs would might want to add asset.
this will output the output_url in the `upload_url` parameter and be made accessible to the release_debian job

> Why not simply use steps.create_release.outputs.upload_url directly?

Could you add an example of your suggestion ?

Wenzel:
cargo release simply bumps the version in Cargo.toml and pushes a new tag
this version tag is the release name, so it gets propagated to the release anyway.

Not sure I understand what you were trying to say though ?

.github/workflows/ci.yml
rageagainsthepc: Sorry, my bad. I was probably thinking of this: https://github.com/Wenzel/libmicrovmi/blob/31f52ba297e07aac3828514382fca2755c8d77c8/.github/workflows/security-audit.yml#L20

.github/workflows/ci.yml
rageagainsthepc: Well, afaik there is no difference between your `step_upload_url` variable and your `create_release` variable. If `step_upload_url` is available from other jobs, `create_release` should be too.

rageagainsthepc:
I wasn't trying to say anything ;) I simply didn't know how cargo release works. (Though I suspected something like this.)

PR #150, opened 2020-12-20
Xen singlestep support

Wenzel:


Wenzel:
This PR brings singlestep support for Xen, and implements the `reply_event` method, which was missing.

I'm still having an issue to listen on multiple VCPU, I only receive events from VCPU 0

src/driver/xen.rs
```diff
@@ -71,18 +77,40 @@ impl Xen {
         let (_ring_page, back_ring, remote_port) = xc
             .monitor_enable(cand_domid)
             .expect("Failed to map event ring page");
-        let xev = XenEventChannel::new(cand_domid, remote_port).unwrap();
 
+        // enable singlestep monitoring
+        // it will only intercept events when explicitely requested using
+        // xc_domain_debug_control()
+        // TODO: call get_vcpu_count()
```
rageagainsthepc, line 84: That comment is obsolete afais, right?

Wenzel: Actually no.
I copy pasted the get_vcpu_count() implementation here, because if i created  the Xen struct below and tried to call get_vcpu_count() afterwards, it had an issue with `xc` being moved above when constructing the Xen struct.

As I didn't have any solution, I used this.

I suggest we should make a function get_vcpu_count(xc: &Xc) and call it from the trait and the Xen driver init function

Wenzel: I've just refactored it.

rageagainsthepc: Hm, I don't see the problem this works fine for me. On a side note, this function has become too large imho, we should split it.

Wenzel: The problem was that I copy pasted a chunk of code, so there was duplication.
I we update the definition of `get_vcpu_count` one day, the other one becomes obselete, etc.

>  This works fine for me. 

Github won't diplay your link: Non-Image content-type returned

rageagainsthepc: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=661acd93eb80b05457b11da953d02b41

Wenzel:
I can confirm that it traps events for all VCPUs

Capture d’écran de 2020-12-20 13-04-54

src/driver/xen.rs
Wenzel: oh.
I don't know I was seeing having a mutable Xen struct as a bad idea.
it's not, since it's restricted on the init function anyway.

I refactored and implemented your changes, thank you

Wenzel:
One issue that needs a fix is that the `singlestep-events` example leave the domain frozen.
The cause should be that there is a singlestep event left in the ring that requires processing, but we already left.

I thought that closing the monitoring and unbinding would clear the queue, but that's not the case apparently.

Looking at Libvmi `xen_events_destroy()`:
https://github.com/libvmi/libvmi/blob/master/libvmi/driver/xen/xen_events.c#L3253

They do a finall pass of listening for events, if any, to clear the queue.
We should do the same

src/driver/xen.rs
```diff
@@ -300,71 +328,106 @@ impl Introspectable for Xen {
...
-                },
-                XenEventType::Msr { msr_type, value } => EventType::Msr { msr_type, value },
-                XenEventType::Breakpoint { insn_len, .. } => {
-                    EventType::Breakpoint { gpa: 0, insn_len }
+            1 => {
+                // event available
+                match self.xev.xenevtchn_pending()? {
+                    -1 => {
+                        // no event channel port is pending
+                        // TODO: Err
```
rageagainsthepc, line 346: Is there a reason why you haven't implemented this yet? Feels to me like it should be done in this PR.

src/driver/xen.rs
```diff
@@ -300,71 +328,106 @@ impl Introspectable for Xen {
...
+                                "Event received for invalid port {}, expected port {}",
+                                pending_event_port, bind_port
+                            );
+                        }
+                        // unmask
+                        self.xev.xenevtchn_unmask(pending_event_port.try_into()?)?;
+                    }
+                };
+                let back_ring_ptr = &mut self.back_ring;
+                match RING_HAS_UNCONSUMED_REQUESTS!(back_ring_ptr) != 0 {
```
rageagainsthepc, line 362: Pattern matching is overkill for a simple boolean check. Also, didn't we come to the conclusion that all those macros could have been simple functions?

src/driver/xen.rs
Wenzel: Because this means creating custom error types, and I'm not feeling confident / knowleadgable enough on the topic to implement this.

src/driver/xen.rs
```diff
@@ -300,71 +328,106 @@ impl Introspectable for Xen {
...
+        };
+        // get the request back
+        let vcpu_index: usize = event.vcpu.try_into()?;
+        let req: vm_event_request_t = mem::replace(&mut self.vec_events[vcpu_index], None).unwrap();
+        let mut rsp: vm_event_response_t =
+            unsafe { mem::MaybeUninit::<vm_event_response_t>::zeroed().assume_init() };
+        rsp.reason = req.reason;
+        rsp.version = VM_EVENT_INTERFACE_VERSION;
+        rsp.vcpu_id = req.vcpu_id;
+        rsp.flags = req.flags & add_flags;
```
rageagainsthepc, line 423:
```suggestion
        let mut rsp = vm_event_response_t {
            reason: req.reason,
            version: VM_EVENT_INTERFACE_VERSION,
            vcpu_id: req.vcpu_id,
            flags: req.flags & add_flags,
            ..Default::default()
        };
```

src/driver/xen.rs
```diff
@@ -387,6 +450,15 @@ impl Introspectable for Xen {
             InterceptType::Breakpoint => {
                 Ok(self.xc.monitor_software_breakpoint(self.domid, enabled)?)
             }
+            InterceptType::Singlestep => {
+                let op: u32 = match enabled {
+                    false => XEN_DOMCTL_DEBUG_OP_SINGLE_STEP_OFF,
+                    true => XEN_DOMCTL_DEBUG_OP_SINGLE_STEP_ON,
+                };
```
rageagainsthepc, line 457:
```suggestion
                let op: u32 = if enabled {
                    XEN_DOMCTL_DEBUG_OP_SINGLE_STEP_ON
                } else {
                    XEN_DOMCTL_DEBUG_OP_SINGLE_STEP_OFF
                };
```

src/driver/xen.rs
rageagainsthepc: Yeah, the overall error handling in this library is not very good. I once tried to refactor it, but got discouraged after realizing how much I would need to change.

rageagainsthepc: At some point we should create a custom error type (or perhaps even multiple, one for each driver) and implement error wrapping instead of error boxing. This is the way. ;)

src/driver/xen.rs
Wenzel: > Also, didn't we come to the conclusion that all those macros could have been simple functions?

Probably, but my focus right now is to make it work.
I'm sure we can improve afterwards.

src/driver/xen.rs
Wenzel: I'm totally in favor of something like this.
I'm simply lacking the time to invest in learning state of the art Rust, and to reflect the changes in libmicrovmi.

But my heart breaks every time I call `unwrap()` in libmicrovmi simply because I'm a noob in error handling.

Wenzel:
the latest commits implemented your fixes @rageagainsthepc 

I also implemented a cleanup of the ring when we close the Xen driver.
And even I can clearly see that we catch and replying to a remaining singlestep event:
Capture d’écran de 2020-12-20 13-58-05


My domain is still frozen.
if you  have ideas.

src/driver/xen.rs
rageagainsthepc: :D

rageagainsthepc:
> My domain is still frozen.
if you have ideas.

The cleanup code looks good to me. :thinking:

Wenzel:
I add debug output in xenevtchn
Capture d’écran de 2020-12-20 15-07-04

Wenzel:
So, latest update.

the domain survives now, but only if it has 1 VCPU.
with 2 VCPU it freezes, which means that the second VCPU is missing a singlestep shutdown API call somehow, or it misses an event notification, idk.

I added a `--count` argument to `singlestep-events` example, to only catch a limited amount of events.

Wenzel:
Here is the `xen-access.c` source that I tweaked
https://gist.github.com/Wenzel/c9fc0ab4f6d4e878e39dcca0fa2a79f4

compile with `gcc xen-access.c -o xen-access -lxenctrl -lxenevtchn`
run with `sudo ./xen-access <dom_id> singlestep`

rageagainsthepc:
>  is the xen-access.c source that I tweaked

What of it? Do you use it as a "control group" to see what we did differently? Does it work without a problem?

Wenzel:
Yes, it is a small example, catching one singlestep and quit, with prints on any Xen API calls.

Yes, it works.

PR #151, opened 2020-12-30
init function is fallible

Wenzel:
This PR should help solve #7 

- dummy driver has been removed
- add `enum-iterator` to iterate over the `DriverType` enum
- init function is now fallible
- all driver init functions are now fallible

Currently when you try to execute an example on Xen, it will panic in the KVM driver, as it is initialized before KVM:
Capture d’écran de 2020-12-30 18-06-15


Wenzel:
@rageagainsthepc i've updated this PR, and removed setting all 3 linux drivers as default features in the build system.
It's not the behavior that I wanted.

I redesigned the init function here, to make it fallible, and to iterate over each compiled driver, to return the first one that succeeds.

I tried to create a Result for the C API interface, it compiles, but cbindgen doesn't generate a valid libmicrovmi header.
Please give me your opinion on this PR when you have time

src/capi.rs
```diff
@@ -44,8 +52,10 @@ pub unsafe extern "C" fn microvmi_init(
                 .expect("Failed to convert DriverInitParam C struct to Rust equivalent"),
         )
     };
-    let driver = init(&safe_domain_name, optional_driver_type, init_option);
-    Box::into_raw(Box::new(driver)) as *mut c_void
+    match init(&safe_domain_name, optional_driver_type, init_option) {
```
rageagainsthepc, line 55: Ok, so cbindgen obviously has problems converting complex enums into C structs, but that shouldn't be surprising because there is no real counterpart for those in C. Maybe a union type wrapping multiple structs. However I would not try to replicate rust error handling in C because idiomatic error handling in C just works differently. I would simply return null in error cases and additionally add an out parameter so we can set error values. The error value could be a simply enum or maybe a struct containing an enum and a string description of the error.

src/lib.rs
```diff
@@ -19,67 +21,57 @@ use driver::kvm::Kvm;
...
-                return create_kvm(domain_name, init_option);
+            // for each possible DriverType
+            for drv_type in DriverType::into_enum_iter() {
+                // try to init
+                match init_driver(domain_name, drv_type, init_option.clone()) {
+                    Ok(driver) => {
+                        return Ok(driver);
+                    }
+                    Err(e) => {
+                        info!("{}", e);
```
rageagainsthepc, line 44: nit: I would not display DriverNotCompiled errors on info log level.

Wenzel:
modifications:

- updated the init C function to take an additional `**const c_char`, to set a string message in case of microvmi init failure
- changed the last MicrovmiError variant display message to display its source

and it works ! :tada: 
Capture d’écran de 2021-01-31 18-04-04


C API
Capture d’écran de 2021-01-31 18-15-26

src/capi.rs
```diff
@@ -23,12 +24,18 @@ pub unsafe extern "C" fn microvmi_envlogger_init() {
...
+///
+/// In case of error, init_error will be allocated with the underlying error message.
+/// The init_error pointer should be freed with rs_cstring_free()
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe extern "C" fn microvmi_init(
     domain_name: *const c_char,
     driver_type: *const DriverType,
     driver_init_option: *const DriverInitParamFFI,
+    init_error: *mut *const c_char,
```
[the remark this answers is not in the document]
Wenzel: if the initialization fails, I don't see what the user can do except log and exit.
of course it would be better to give more meaningful info to the C world, what type of error is reported for example.

libvmi fills an enum to try to convey the cause of the failure.
https://github.com/libvmi/libvmi/blob/master/libvmi/libvmi.h#L93

libmicrovmi is already doing better by providing a precise error string, and not a generic one i'd say.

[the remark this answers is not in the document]
Wenzel: f that's okay for you, you can resolve this conversation.

src/capi.rs
```diff
@@ -17,18 +18,24 @@ pub enum DriverInitParamFFI {
...
     env_logger::init();
 }
 
-#[allow(clippy::missing_safety_doc)]
+/// Entrypoint for libmicrovmi
+/// Initializes a specific driver, or all drivers compiled and returns the first one that succeeded
+///
+/// In case of error, init_error will be allocated with the underlying error message.
+/// The init_error pointer should be freed with rs_cstring_free()
+/// # Safety
```
rageagainsthepc, line 32: Is that an empty Safety section? Why?

Wenzel: because there is no additional safety comment to add, but clippy will throw a warning without this section.
maybe i should have disabled the clippy warning then ?

rageagainsthepc: Well, the safety section should document the caller's responsibilities in order to uphold safety guarantees, so you should move line 31 into this section, for example.

PR #152, opened 2020-12-31
Refactor error handling

rageagainsthepc:


rageagainsthepc:
I'm not quite happy yet, error handling has become awkward because I'm wrapping everything into an intermediate error.

Wenzel:
Hi @rageagainsthepc,
thanks for the draft.

When I started to review your PR, i googled some questions I had and found a good talk about error handling at RustConf 2020:
https://www.youtube.com/watch?v=rAF8mLI0naQ

I answered my questions:
# 1 - should we use Box<CustomErrorType> or CustomErrorType ?

According to https://youtu.be/rAF8mLI0naQ?t=1401
If the error becomes large, it should be better to allocated it on the heap.
Let's follow this behavior.

# 2 - API stability

According to https://youtu.be/rAF8mLI0naQ?t=1360
We should add `#[non_exhaustive]` to keep the Error enum stable even if we add new error types afterwards

# 3 - returning a error type vs returning an error trait

Returning
~~~Rust
-> Result<(), Box<dyn DriverError>>
~~~
with `DriverError` is more or less the same as returning the `std::error::Error` trait in my opinion.

Also, as we are a library, we need to return flexible and precise errors to the user.
https://youtu.be/rAF8mLI0naQ?t=1657

with `Box<dyn DriverError>`, the user has to downcast dynamically retrieve the associated error.
I would like to try to use an enum as much as possible

Thanks !

rageagainsthepc:
> with DriverError is more or less the same as returning the std::error::Error trait in my opinion.

Defining this trait forces us to implement a custom error type that derives from this trait.

> with Box<dyn DriverError>, the user has to downcast dynamically retrieve the associated error.

Yeah, that's definitely not nice, but it is also the status quo and I don't have a nice way around this currently.
Does the talk address the problem with errors in traits by any chance?

Wenzel:
After posting our problems on the Rust community forum:
https://users.rust-lang.org/t/error-handling-in-library-enum-or-trait-type/53750/4

I think we should go with an Error associated type for the `Introspectable` trait, and have each backend define an enum, deriving from `thiserror`, providing all the possible backend specific errors.

Also, having an error trait `DriverError` is not needed, and we should use the `std::error::Error` trait itself

Wenzel:
So according to the last solution posted, and with an example of implementation here:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=0b6ef3b91cfeaa75a74278f297044930

Using a builder approach would help us to manage errors.

The last thing needs a discussion is that fact that the `DriverError` si represented as a enum like this:
~~~rust
enum DriverError {
    Xen(XenDriverError),
    HyperV(HyperVDriverError),
}
~~~

This design exclude the possibility reusability of the crate by prevents a third party importing the trait and implementing his own driver.

If you have an idea @rageagainsthepc ?

rageagainsthepc:
1. We choose a monolithic approach where we say new drivers should always be merged into this repo. This means that we would need to be willing to maintain those drivers going forward.
2. We implement an opaque error type and perhaps provide convenience methods for downcasting errors.

I'm conflicted but my gut feeling leans towards option 1.

Wenzel:
> I'm conflicted but my gut feeling leans towards option 1.

I also tend to choose option 1. It's not the most flexible, but it's straightforward to implement.
When libmicrovmi will (eventually) become popular enough that people would want to develop their own drivers, we can rethink that implemention.

till then, we have a long road ahead :)

Wenzel:
Now, as you said, next step would be to implement the builder pattern.
Regarding the naming, as the rust book is saying, i would like to attempt to remove the `Builder` name.
Like a `Process` is created by a `Command` (and `Command` implements the builder)
I would like to suggest that a `Xen/KVM/VBoxDriver` (`Box<dyn Introspectable>` in the end) is created by an `IntrospectableVM`.

~~~rust
    let driver = IntrospectableVM::new("windows10-20h2")
        .driver_type(XenDriver)
        .init();
~~~

It makes sense that, from a Virtual Machine that has the capability of being introspected, through the means of a yet unknown hypervisor (the driver_type is defined in the next call after all), we are instanciating a driver.


Also, the `init()` function should look like this:
~~~rust
    fn init(self) -> Result<Box<dyn Introspectable<DriverError = D::DriverError>>, MicrovmiError> {
        let IntrospectableBuilder {
            domain_name,
            init_option,
            driver_type,
        } = self;

        // copy the desired behavior of the original init function
        // driver_type needs to be an Option(DriverType)
        drv = match self.driver_type {
              Some() => // match on the specified driver type, and try to instantiate the driver with ::new()
              None => // iterate on all possible drivers, and if all fails, return a MicrovmiError
        }

        Ok(drv)
    }
~~~


What do you think ?

rageagainsthepc:
Looks good to me on first glance, will need to play around with it a little though. There are still a few things that are very much unclear to me.

Wenzel:
> There are still a few things that are very much unclear to me.

me too, for example, I'm unsure if having an `Option(DriverType)` is actually needed, since the option is expressed by the presence of `AnyDriver` struct as `driver_type` field, or not ?

But i suppose the best way is to head right into the implementation

Wenzel:
I suppose you can close this PR ?

PR #153, opened 2021-01-04
Better error handling

Wenzel:
This PR is a draft to implement precise error types in libmicrovmi.

It tries to improve @rageagainsthepc 's https://github.com/Wenzel/libmicrovmi/pull/152, by removing the DriverError trait and returning only clearly defined enums.

All the trait function definition have been moved to returning `Result<T, DriverError>`
Contrary to what I said https://github.com/Wenzel/libmicrovmi/pull/152#issuecomment-753631651 about boxing errors, i attempted here to remove any boxing, since:
- there is no trait anymore, so the size is known
- we don't know yet if we actually really need to box our errors because of the stack size with large error types. this is something i would like to discuss with members of the rust community
- boxing makes the code a bit less readable

the new trait definition has been implemented on:
- Dummy
- VirtualBox

Furthermore, the `init` function for these 2 drivers is now fallible

new error types:
- `MicrovmiError`
- `DriverError`
- `DummyDriverError` and `VirtualBoxDriverError`

Last thing, I will switch the error types to be `#[non_exhaustive]` to better understand all the implications when updating an driver error type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ hyper-v = ["winapi", "widestring", "ntapi", "vid-sys"]
 [dependencies]
 log = "0.4.8"
 env_logger = "0.7.1"
+thiserror = "1.0.23"
 libc = { version = "0.2.58", optional = true }
 xenctrl = { version = "0.4.2", optional = true }
 xenstore-rs = { version = "0.3.0", optional = true }
--- a/src/api.rs
+++ b/src/api.rs
@@ -3,6 +3,8 @@ use std::ffi::{CStr, IntoStringError};
 
 use crate::capi::DriverInitParamFFI;
 use crate::driver::dummy::DummyDriverError;
+#[cfg(feature = "virtualbox")]
+use crate::driver::virtualbox::VirtualBoxDriverError;
 
 bitflags! {
     pub struct Access: u32 {
@@ -186,6 +188,9 @@ pub enum MicrovmiError {
 pub enum DriverError {
     #[error(transparent)]
     Dummy(#[from] DummyDriverError),
+    #[cfg(feature = "virtualbox")]
+    #[error(transparent)]
+    VirtualBox(#[from] VirtualBoxDriverError),
 }
 
 pub trait Introspectable {
--- a/src/api.rs
+++ b/src/api.rs
@@ -177,7 +177,10 @@ pub const PAGE_SHIFT: u32 = 12;
 pub const PAGE_SIZE: u32 = 4096;
 
 #[derive(thiserror::Error, Debug)]
-pub enum MicrovmiError {}
+pub enum MicrovmiError {
+    #[error(transparent)]
+    DriverError(#[from] DriverError),
+}
 
 #[derive(thiserror::Error, Debug)]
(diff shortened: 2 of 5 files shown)


Cargo.toml
```diff
@@ -27,6 +27,7 @@ hyper-v = ["winapi", "widestring", "ntapi", "vid-sys"]
 [dependencies]
 log = "0.4.8"
 env_logger = "0.7.1"
+thiserror = "1.0.23"
```
rageagainsthepc, line 30: nit: I would (generally) not pin the patch version. The `thiserror` maintainers also suggest simply using `1.0`.

src/api.rs
```diff
@@ -176,10 +178,25 @@ pub enum Registers {
 pub const PAGE_SHIFT: u32 = 12;
 pub const PAGE_SIZE: u32 = 4096;
 
+#[derive(thiserror::Error, Debug)]
+pub enum MicrovmiError {
+    #[error(transparent)]
```
rageagainsthepc, line 183: So, in the talk you linked in my PR this behavior is essentially regarded as a bug (if I'm not mistaken) because it messes with proper error reporting.

src/api.rs
```diff
@@ -176,10 +178,25 @@ pub enum Registers {
...
+    DriverError(#[from] DriverError),
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum DriverError {
```
rageagainsthepc, line 188: This kinda bums me out. We have to wrap errors from each driver that implements the `Introspectable` trait which essentially breaks abstraction. :/

src/driver/virtualbox.rs
```diff
@@ -3,103 +3,216 @@ use std::error::Error;
...
         // no API to fetch VCPU count, hardcode to 1 for now
         Ok(1)
     }
 
-    fn read_physical(&self, paddr: u64, buf: &mut [u8]) -> Result<(), Box<dyn Error>> {
-        self.fdp.read_physical_memory(paddr, buf)
+    fn read_physical(&self, paddr: u64, buf: &mut [u8]) -> Result<(), DriverError> {
+        self.fdp
+            .read_physical_memory(paddr, buf)
+            .map_err(|err| VirtualBoxDriverError::OtherError(err).into())
```
rageagainsthepc, line 42: This is a case where `snafu` could shine with its `context` method.

src/driver/virtualbox.rs
```diff
@@ -3,103 +3,216 @@ use std::error::Error;
...
 use crate::api::{
-    DriverInitParam, Introspectable, Registers, SegmentReg, SystemTableReg, X86Registers,
+    DriverError, DriverInitParam, Introspectable, Registers, SegmentReg, SystemTableReg,
+    X86Registers,
 };
 
+#[derive(thiserror::Error, Debug)]
+pub enum VirtualBoxDriverError {
+    #[error(transparent)]
+    OtherError(#[from] Box<dyn Error>),
```
rageagainsthepc, line 13: My idea was to first get the error handling in our dependencies right, so we don't have to do things like this, for example. (Also, it's a lot less controversial than error handling in this crate).

PR #154, opened 2021-01-08
Add gitpod

rageagainsthepc:
Gitpod provides Web IDEs which I think work already really well. I can definitely see myself using those instead of a local IDE (unless I have to actually run the code of course). It should also be handy for new people who want to explore the repo. However, I've enabled experimental features in order to be able to use VSCode instead of Theia because the rust-analyzer plugin does not work with the latter.

Edit: I should mention that experimental features are an account setting unfortunately.

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,22 @@
+image:
+  file: .gitpod.Dockerfile
+
+tasks:
+  - init: cargo build --features kvm,virtualbox,xen
+
+vscode:
+  extensions:
+    # Theia does not support rust-analyzer, therefore this plugin list targets vscode instead. 
+    - editorconfig.editorconfig
+    - rust-lang.rust
+    - matklad.rust-analyzer
+    - eamodio.gitlens
+    - serayuzgur.crates
+    - belfz.search-crates-io
+    - bungcip.better-toml
+
+github:
+  prebuilds:
+    branches: true
+    addCheck: false
+    addComment: true
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,25 @@
+FROM gitpod/workspace-full
+
+RUN sudo apt-get update \
+    && sudo apt-get install -y --no-install-recommends clang libxen-dev \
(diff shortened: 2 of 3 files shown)


.gitpod.yml
```diff
@@ -0,0 +1,21 @@
+image:
+  file: .gitpod.Dockerfile
+
+tasks:
+  - init: cargo build --features kvm,virtualbox,xen
+
+vscode:
+  extensions:
+    - editorconfig.editorconfig
```
Wenzel, line 9: Can you put a comment on these extensions and why they are needed ?

.gitpod.yml
```diff
@@ -0,0 +1,21 @@
...
+    - rust-lang.rust
+    - matklad.rust-analyzer
+    - eamodio.gitlens
+    - serayuzgur.crates
+    - belfz.search-crates-io
+    - bungcip.better-toml
+
+github:
+  prebuilds:
+    branches: true
```
Wenzel, line 19: so [`addBadge`](https://www.gitpod.io/docs/prebuilds/#badge) modifies the description a of PR ?
https://www.gitpod.io/docs/prebuilds/#badge
this is really invasive

We can use https://www.gitpod.io/docs/prebuilds/#comment instead to display a button to open Gitpod for a given PR.

Wenzel (changes requested):

.gitpod.yml
rageagainsthepc: Are you sure you want me to do that? Seems a bit redundant provided most of my comments would probably be paraphrasing the first sentence in the corresponding plugin summary. That said I have no problem with adding comments, just wanted to double check.

Wenzel: The thing is that replaced the default ide for vscode, because rust analyzer support issues.
So at least a comment could be placed here to explain why we are using experimental extensions.

That's all i meant.

Wenzel (commented):
See my comment

.gitpod.yml
rageagainsthepc: Oh yeah, that makes sense. :D Good call 👍

Wenzel (approved):

PR #155, opened 2021-01-08
Fixes

rageagainsthepc:


c_examples/regs-dump.c
```diff
@@ -14,17 +14,17 @@ void read_registers(void* driver, const char* vm_name) {
     Registers regs;
     memset(&regs, 0, sizeof(regs));
     if (microvmi_read_registers(driver, 0, &regs)) {
-        printf("rax: 0x%" PRIx64 "\n", regs.x86._0.rax);
```
rageagainsthepc, line 17: How could this even happen? Shouldn't this have been cought by the CI when the enum was removed?

Wenzel: I believe this happened because we install cbindgen from the latest version, and the last one broke this enum, so I had to fix it.
best way to fix this is to pin cbindgen crate both in CI and c_examples/README.md

rageagainsthepc: Ok, I see. I would actually prefer not to pin the version. Unless of course it becomes too much of a hassle to maintain compatibility withh newer releases of cbindgen.

rageagainsthepc: I mean, this would have been caught by our CI if we would run it periodically. Also, I think the result is an improvement (although just a small one).

rageagainsthepc:
@Wenzel  Can we merge this?

c_examples/regs-dump.c
Wenzel: > I would actually prefer not to pin the version. 

If we don't pin, we let our users try our library and eventually fail to compile the `c_examples`, unbeknownst to us.

> I mean, this would have been caught by our CI if we would run it periodically.

Running the CI periodically is a solution to the wrong problem.
we need to ping cbindgen, and periodically update it

rageagainsthepc:
> Running the CI periodically is a solution to the wrong problem.
we need to pin cbindgen, and periodically update it

In my personal experience version pinning just leads to depecrated and potentially insecure libraries. I believe our dependency checker CI job has already opened one or two issues about such cases which we have ignored so far. Of course that does not mean we should risk breaking the library ever so often but we should at least be mindful of pinning too much.
In this case we have no control over which version of `cbindgen` the user chooses to install because it's a tool, not a dependency. So it's best to assume that the user will choose the most recent version and we should try to be compatible to that.

Cargo.toml
```diff
@@ -25,27 +25,31 @@ hyper-v = ["winapi", "widestring", "ntapi", "vid-sys"]
...
-colored = "1.9.3"
-mockall = "0.7.1"
+ctrlc = "3.1"
+clap = "2.33"
+colored = "2.0"
+mockall = "0.9"
 test-case = "1.0.0"
+
+# Wait for test-case 1.0.1 release
+[patch.crates-io]
```
Wenzel, line 54: let's wait for the official release on crates.io,
I don't want to merge a patched crate based on a git repo.

rageagainsthepc: Sure, but it's uncertain when that will be and until then the CI is broken. You could try to pin `syn` to an older version instead of patching `test-case`, but I tried and I was not successful with that.

Wenzel: Then just pin it to a specific commit with `ver`.
any changes on test-case master branch can brake our CI with this.

Wenzel: Edit: sorry `rev`, not `ver`

rageagainsthepc: Never mind, turns out they were quick about the release. ;)

PR #160, opened 2021-01-29
Driver error enum

Wenzel:
This PR adds driver specific error types based on `thiserror` for Xen and KVM

regarding virtualbox, the FDP crate returns `Box<dyn Error>` for every function, so implementing a better handling for this driver will start with FDP library.

### DIFF
--- a/src/driver/xen.rs
+++ b/src/driver/xen.rs
@@ -5,15 +5,19 @@ use crate::api::{
 use libc::{PROT_READ, PROT_WRITE};
 use nix::poll::PollFlags;
 use nix::poll::{poll, PollFd};
+use std::convert::Infallible;
 use std::convert::TryInto;
 use std::error::Error;
+use std::io::Error as IoError;
 use std::io::ErrorKind;
 use std::mem;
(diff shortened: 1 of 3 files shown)


src/driver/xen.rs
```diff
@@ -29,6 +33,24 @@ pub struct Xen {
     back_ring: vm_event_back_ring,
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum XenDriverError {
+    #[error("event version mismatch: {0} <-> {1}")]
+    EventVersionMismatch(u32, u32),
+    #[error(transparent)]
```
rageagainsthepc, line 40: IIRC this duplicates the error message from the source error and therefore makes walking error source chains awkward.

Wenzel: So we should replace `#[error(transparent)]` by `#[error("defined error message")]` ?

Wenzel: okay, done

Wenzel:
Merging this to work on #151

PR #165, opened 2021-02-02
pyapi: generate basic python bindings with PyO3

### LINKED ISSUE #162: libmicrovmi python bindings
Create Python bindings to call libmicrovmi from Python

TODO (updated):

- [x] decide how to integrate this new library in our build system
- [x] expose `DriverType` and `DriverInitParam`
- [x] expose `read_physical`
- [ ] package and upload to PyPI
- [x] update `setup.py` to accept `--features` parameter, and set features in `RustExtension`
- [x] create new Dockerfile based on `quay.io/pypa/manylinux2014_x86_64` and install all the drivers required development headers
- [x] update `build-wheels.sh` to accept `--features` and pass them to `setup.py`
- [x] read all relevant fields of `python/Cargo.toml` to fill `setup.py` metadata

We should use the [`rust-cpython` project](https://github.com/dgrunwald/rust-cpython#example-library-with-python-bindings)

rust-cpython will help us to generate a Python extension module, wrapping the libmicrovmi Rust API.

# how to integrate this new library

1. create a separate repository
2. use cargo workspace to manage both Rust library and Python extension in the same git repository, in tandem.
3. maybe generate both with the same cargo.toml (I doubt this is possible)

@rageagainsthepc if you have an opinion

Wenzel:
Generate (basic) python bindings to initialize libmicromi from Python, based on [`PyO3`](https://github.com/PyO3/pyo3)
fix #162 

Capture d’écran de 2021-02-02 16-21-12


Wenzel:
So I managed to expose the DriverType and DriverInitParam types to Python.

It might not be the best way, but it works
Capture d’écran de 2021-02-02 22-14-44

There is an opened [issue](https://github.com/PyO3/pyo3/issues/417) on PyO3 to discuss how they should implement and expose Rust enum to Python.
Maybe in the future we could remove this code and let it managed by PyO3.

Wenzel:
Note: I had to remove the conditional compilation in DriverType:
https://github.com/Wenzel/libmicrovmi/pull/165/files#diff-adea2445327e39eb215de697b440dbbc02135357380fb612a10a9e659129a46bL25

But this is what we wanted with #151 anyway, so i will rebase once this other PR merged.

python/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+xen = ["microvmi/xen"]
+# KVM driver
+kvm = ["microvmi/kvm"]
+# VirtualBox driver
+virtualbox = ["microvmi/virtualbox"]
+# Hyper-v driver
+hyper-v = ["microvmi/hyper-v"]
+
+[dependencies]
+log = "0.4.8"
```
rageagainsthepc, line 30: nit: Don't pin the patch version

python/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+env_logger = "0.8"
+thiserror = "1.0"
+microvmi = { path = "../" }
+
+[dependencies.pyo3]
+version = "0.13.1"
```
rageagainsthepc, line 36: nit: Same here, don't pin the patch version

python/requirements.txt
```diff
@@ -0,0 +1,4 @@
+setuptools-rust
+setuptools
+wheel
+toml==0.10.2
```
rageagainsthepc, line 4: Why the exact version pinning here?

python/setup.py
```diff
@@ -0,0 +1,44 @@
...
+with open(str(CUR_DIR / "README.md"), "r", encoding="utf-8") as fh:
+    long_description = fh.read()
+
+with open(str(CUR_DIR / "requirements.txt")) as f:
+    requirements = f.read().splitlines()
+
+setup(
+    name="microvmi",
+    version=cargo_version,
+    author="Mathieu Tarral",
```
rageagainsthepc, line 27: There is a lot of duplicated information here. Can't you somehow funnel the information from Cargo.toml to setuptools?

Wenzel: with the `toml` python package, we are already forward the version information.
good idea, let's do it for the rest of the setuptools fields

python/src/errors.rs
```diff
@@ -0,0 +1,29 @@
...
+
+use pyo3::exceptions::PyValueError;
+use pyo3::PyErr;
+
+use microvmi::errors::MicrovmiError;
+
+// create a NewType for MicrovmiError, since we cannot implement
+// From trait on external types
+#[derive(thiserror::Error, Debug)]
+pub enum PyMicrovmiError {
```
rageagainsthepc, line 12: This sucks. Remind me again why we need a sparate workspace? :/

Wenzel: because we need to create another `dylib`, and it would have clashed with the C library.
And building an hybrid C library for both C API and Python C extension means any C app would have needed to install and link with python-dev.

python/requirements.txt
Wenzel: the rule is that you should always pin your dependencies in requirements.txt, except for `wheel`, because you always need the latest version.
as the readme from setuptools-rust was explicitely not pinning these 3 dependencies, i did the same

python/Cargo.toml
Wenzel: I don't understand, this is not a patch section ?
what do you mean by patch version ?

python/Cargo.toml
Wenzel: pyo3 is bit sensitive, and i would like to keep the patch version pinned
https://pyo3.rs/v0.13.1/changelog.html

Wenzel:
Capture d’écran de 2021-02-05 09-26-09
library init works (here on Xen), but `read_physical` implementation returns an empty List (?)

python/src/errors.rs
rageagainsthepc: Ok, I also don't have a better idea right now, so ¯\_(ツ)_/¯

Wenzel:
Ok, so, this implementation is not working:
~~~rust
        let mut buffer = Vec::with_capacity(size);
        self.driver.read_physical(paddr, &mut buffer).map_err(PyMicrovmiError::from)?;
        debug!("buffer = {:?}", buffer);
~~~

but this one is:
~~~rust
        let mut buffer2: [u8; 1024] = [0; 1024];
        self.driver.read_physical(paddr, &mut buffer2).map_err(PyMicrovmiError::from)?;
        debug!("buffer2 = {:?}", buffer2);
~~~

Why I can't manage to fill my vector and return it ?

Wenzel:
Ok, so the catch was that `Vec::with_capacity` was creating an empty vec, so the array had a size of `0` :clown_face: 

This is fixed now.

Wenzel:
I added a `mem-dump.py`, with a nice Rich progress bar.
The performance results are the following:
`mem-dump.rs`: 27 sec
`mem-dump.py`: 2min 57 sec

I suppose there are a lot of uncessary copies in the Python version somehow

Wenzel:
I'm trying to update the implementation to return a `PyBytes` type, and possibly avoid a conversion:

~~~rust
    fn read_physical(&self, py: Python, paddr: u64, size: usize) -> PyResult<&PyBytes> {
        let pybuffer = PyBytes::new_with(py, size, |mut buffer| {
            self.driver
                .read_physical(paddr, &mut buffer)
                .map_err(PyMicrovmiError::from)?;
            Ok(())
        })?;

        Ok(pybuffer)
    }
~~~

but it doesn't compile, lifetime issue:
~~~
error[E0623]: lifetime mismatch
   --> python/src/lib.rs:158:9
    |
150 |     fn read_physical(&self, py: Python, paddr: u64, size: usize) -> PyResult<&PyBytes> {
    |                                 ------                              ------------------
    |                                 |
    |                                 this parameter and the return type are declared with different lifetimes...
...
158 |         Ok(pybuffer)
    |         ^^^^^^^^^^^^ ...but data from `py` is returned here

error: aborting due to previous error

For more information about this error, try `rustc --explain E0623`.
error: could not compile `pymicrovmi`
~~~

Wenzel:
I got a reply from PyO3 Gitter Lobby, and I managed to compile the improved version based on `&PyBytes`.
The performance is much better !

~~~rust
(venv) vagrant@ubuntu2004:/vagrant/libmicrovmi/python$ time sudo -E ../venv/bin/python examples/mem-dump.py winxp
Dumping physical memory on winxp until 0xFF000000 to winxp.dump

real	0m41.720s
~~~

vs rust
~~~
(venv) vagrant@ubuntu2004:/vagrant/libmicrovmi$ time sudo ~/.cargo/bin/cargo run --features xen --release --example mem-dump winxp
    Finished release [optimized] target(s) in 0.37s
     Running `target/release/examples/mem-dump winxp`
pausing the VM
Max address @ff000000
Dumping physical memory to winxp.dump
resuming the VM

real	0m26.905s
~~~

This means *~37%* performance gap between Rust and Python here, which is not excessive :)

Wenzel:
@rageagainsthepc i believe this PR is ready for review

python/setup.py
rageagainsthepc: Did you forget about this?

Wenzel: i could parse the Cargo author field to fill this one

python/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+    ) -> PyResult<Self> {
+        info!("Microvmi Python init");
+        // convert Python DriverType to rust API DriverType
+        debug!(
+            "Microvmi Python init driver_type: {:?}, init_param: {:?}",
+            driver_type, init_param
+        );
+        let rust_driver_type: Option<rapi::DriverType> = if let Some(drv_type) = driver_type {
+            Some(match drv_type {
+                DriverType::HYPERV => Ok(rapi::DriverType::HyperV),
```
rageagainsthepc, line 108: nit: I would usually favor implementing TryFrom for `DriverType`, but since this probably the only place where it would be used I don't have any strong opionions about it.

python/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+        driver_type: Option<u32>,
+        init_param: Option<DriverInitParam>,
+    ) -> PyResult<Self> {
+        info!("Microvmi Python init");
+        // convert Python DriverType to rust API DriverType
+        debug!(
+            "Microvmi Python init driver_type: {:?}, init_param: {:?}",
+            driver_type, init_param
+        );
+        let rust_driver_type: Option<rapi::DriverType> = if let Some(drv_type) = driver_type {
```
rageagainsthepc, line 106: IIRC using `map` on the option type would be better than using `if else`

python/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+                _ => Err(PyValueError::new_err(format!(
+                    "Invalid value for DriverType: {}",
+                    drv_type
+                ))),
+            }?)
+        } else {
+            None
+        };
+        // convert Python DriverInitParam to rust API DriverinitParam
+        let rust_init_param: Option<rapi::DriverInitParam> = if let Some(param) = init_param {
```
rageagainsthepc, line 121: same here

python/src/lib.rs
```diff
@@ -0,0 +1,182 @@
...
+            Some(
+                #[allow(unreachable_patterns)]
```
rageagainsthepc, line 123: Why this?

Wenzel: because clippy would complain that since there was only one pattern match, it could be replaced by the matched body itself

python/src/lib.rs
Wenzel: I tried, but had an issue with the `?` operator
~~~rust
        let rust_driver_type: Option<rapi::DriverType> = driver_type.map(|drv_type| match drv_type {
                DriverType::HYPERV => Ok(rapi::DriverType::HyperV),
                DriverType::KVM => Ok(rapi::DriverType::KVM),
                DriverType::VIRTUALBOX => Ok(rapi::DriverType::VirtualBox),
                DriverType::XEN => Ok(rapi::DriverType::Xen),
                _ => Err(PyValueError::new_err(format!(
                    "Invalid value for DriverType: {}",
                    drv_type
                ))),
            }?);
~~~

~~~
error[E0277]: the `?` operator can only be used in a closure that returns `Result` or `Option` (or another type that implements `Try`)
   --> python/src/lib.rs:106:85
    |
106 |            let rust_driver_type: Option<rapi::DriverType> = driver_type.map(|drv_type| match drv_type {
    |   __________________________________________________________________________-__________^
    |  |__________________________________________________________________________|
    | ||
107 | ||                 DriverType::HYPERV => Ok(rapi::DriverType::HyperV),
108 | ||                 DriverType::KVM => Ok(rapi::DriverType::KVM),
109 | ||                 DriverType::VIRTUALBOX => Ok(rapi::DriverType::VirtualBox),
...   ||
114 | ||                 ))),
115 | ||             }?);
    | ||              ^
    | ||______________|
    | |_______________cannot use the `?` operator in a closure that returns `microvmi::api::DriverType`
    |                 this function should return `Result` or `Option` to accept `?`
    |
    = help: the trait `Try` is not implemented for `microvmi::api::DriverType`
    = note: required by `from_error`
~~~

:(

Wenzel:
I fixed the mem-dump.py implementation, we need to write a 0 page when the page doesn't exist.
Capture d’écran de 2021-02-09 11-23-36

python/src/lib.rs
```diff
@@ -0,0 +1,195 @@
...
+                DriverType::VIRTUALBOX => Ok(rapi::DriverType::VirtualBox),
+                DriverType::XEN => Ok(rapi::DriverType::Xen),
+                _ => Err(PyValueError::new_err(format!(
+                    "Invalid value for DriverType: {}",
+                    drv_type
+                ))),
+            }?)
+        } else {
+            None
+        };
```
rageagainsthepc, line 119:
```suggestion
let rust_driver_type = driver_type.map(|drv_type| match drv_type {
                a if a == DriverType::HYPERV => Ok(rapi::DriverType::HyperV),
                a if a == DriverType::KVM => Ok(rapi::DriverType::KVM),
                a if a == DriverType::VIRTUALBOX => Ok(rapi::DriverType::VirtualBox),
                a if a == DriverType::XEN => Ok(rapi::DriverType::Xen),
                _ => Err(PyValueError::new_err(format!(
                    "Invalid value for DriverType: {}",
                    drv_type
                )))}).transpose()?;
```

python/src/lib.rs
rageagainsthepc: see my suggestion

python/src/lib.rs
rageagainsthepc: Shouldn't you make `DriverInitParam` non_exhaustive instead?

Wenzel:
a couple of updates here:
- a fix in xen 's `read_physical` (again) where the `cur_paddr` wasn't updated, leading to reading the same chunk of bytes over and over...
- an attempt to expose the amount of `bytes_read` in `read_physical`. I will see how it helps me to integrate the read with Volatility.

python/src/lib.rs
Wenzel: non exhaustive is more dangerous.
it means that when we add a new enum, this match arm will still compile, even though we forgot to handle the new enum value, so it's not the right solution here

rageagainsthepc: Right, but this also means that adding values to `DriverInitParam` (I mean the one we expose to our users) is always a breaking change. But if that's okay with you we can leave it like this.

Wenzel: SInce the python library recompiles libmicrovmi and distributes it, this is not an issue.

c_examples/mem-dump.c
```diff
@@ -24,7 +24,8 @@ void dump_memory(void* driver, const char* vm_name) {
     uint8_t buffer[PAGE_SIZE];
     for (int i = 0; i <= max_address / PAGE_SIZE; i++) {
         memset(buffer, 0, PAGE_SIZE);
-        if (microvmi_read_physical(driver, i * PAGE_SIZE, buffer, PAGE_SIZE)) {
+	    uint64_t bytes_read = 0;
```
rageagainsthepc, line 27: Indentation 😱 😱 😱

Wenzel: Thanks.
I edited the file with vi in vagrant, and then with Intellij, and the IDE didn't complain about it.

Wenzel:
Ready for review

python/examples/mem-dump.py
```diff
@@ -0,0 +1,73 @@
...
+
+    try:
+        vm_name = args.vm_name
+        output_file = args.output_file
+        dump_mem(vm_name, output_file)
+    except KeyboardInterrupt:
+        logging.critical("Cancelled by CTRL-C")
+
+
+main()
```
rageagainsthepc, line 73:
```suggestion
if __name__ == "__main__":
    main()
```

Wenzel: Fixed !

rageagainsthepc:
Right now the gitpod image does not work because you've pinned a rev which does not exist anymore. Why did you want to check out a different branch than we use in the CI anyways?

src/capi.rs
```diff
@@ -93,10 +93,23 @@ pub unsafe extern "C" fn microvmi_read_physical(
     physical_address: uint64_t,
     buffer: *mut uint8_t,
     size: size_t,
+    bytes_read: *mut uint64_t,
 ) -> bool {
     let driver = get_driver_mut_ptr(context);
+    let mut ignored_bytes_read: u64 = 0;
+    // if bytes_read param is NULL, just ignore the bytes_read in this local variable
+    let bytes_read_ptr = if bytes_read.is_null() {
```
rageagainsthepc, line 101: nit: This is a little awkward from a design perspective. Why not using a local var for the `read_physical` call and assigning the value to `bytes_read` afterwards (if it is not null)? This way you have one pointer var less and you don't need to conditionally assign pointers to other pointers.

src/api.rs
```diff
@@ -186,8 +186,14 @@ pub trait Introspectable {
     ///
     /// * 'paddr' - the physical address to read from
     /// * 'buf' - the data read from memory
+    /// * 'bytes_read' - the number of bytes read
     ///
-    fn read_physical(&self, _paddr: u64, _buf: &mut [u8]) -> Result<(), Box<dyn Error>> {
+    fn read_physical(
```
rageagainsthepc, line 191: I get why the buffer is passed as an argument, but why `bytes_read` as well? Wouldn't a return value be nicer and also more in line with other low level APIs that read stuff into a buffer?

src/capi.rs
Wenzel: You're right, i refactored the code, it's much clearer now.

src/api.rs
Wenzel: It's true that we could have returned the amount of `bytes_read` using the `Result<>` `Ok(bytes_read)`.

Although, in the case of an Error, the caller loose this information, and he doesn't know how much bytes have already been read successfully.
So if you do a big read operation, from the start of the physicial memory, you will invevitably get an error because of unmapped pages, and you get an unusable buffer after the call.

That's why i kept it there.
But with https://github.com/Wenzel/libmicrovmi/issues/175 we can explore and redesign this concept, given that std lib `Read` trait's `read()` function is returning the amount of bytes_read as an Ok Result.

Wenzel: This also mimics libvmi:
https://github.com/libvmi/libvmi/blob/master/libvmi/read.c#L284

if a page is found to be NULL, a jump is taken to the [done label](https://github.com/libvmi/libvmi/blob/master/libvmi/read.c#L303), updating the `bytes_read` according while returning a failure status

rageagainsthepc: 👍

src/capi.rs
```diff
@@ -93,11 +93,23 @@ pub unsafe extern "C" fn microvmi_read_physical(
...
+    let res = (*driver)
+        .read_physical(
+            physical_address,
+            slice::from_raw_parts_mut(buffer, size),
+            &mut bytes_read_local,
+        )
+        .is_ok();
+    // update bytes_read if not NULL
+    if !bytes_read.is_null() {
+        *bytes_read.as_mut().unwrap() = bytes_read_local;
```
rageagainsthepc, line 110:
```suggestion
        bytes_read.write(bytes_read_local);
```

src/capi.rs
```diff
@@ -93,11 +93,23 @@ pub unsafe extern "C" fn microvmi_read_physical(
...
     buffer: *mut uint8_t,
     size: size_t,
+    bytes_read: *mut uint64_t,
 ) -> bool {
     let driver = get_driver_mut_ptr(context);
-    (*driver)
-        .read_physical(physical_address, slice::from_raw_parts_mut(buffer, size))
-        .is_ok()
+
+    let mut bytes_read_local: u64 = 0;
```
rageagainsthepc, line 100:
```suggestion
    let mut bytes_read_local = 0;
```

Wenzel: done

src/capi.rs
Wenzel: nice catch !

rageagainsthepc (approved):
LGTM

Wenzel:
I fixed the `read` implementation of `PaddedPhysicalMemoryIO`: it was supposed to update the position after each read.
I also updated `mem-dump.py` to use the padded memory, and this produces a valid memory dump.

however, when using volatility directly on live memory, it fails:
Capture d’écran de 2021-03-16 14-22-21

Since the memory dump produced is valid for volatility, the Xen driver is not to blame.

But this implementation of `PaddedPhysicalMemoryIO` works perfectly fine on KVM.
What the heck

Wenzel:
I can't really understand what's happening in volatility here
1. pslist on winxp xen domain with live memory
2. dump winxp xen domain live memory volatility layer to a dump file
3. pslist works on the layer dump file

:crying_cat_face: 

Capture d’écran de 2021-03-16 14-48-26

Wenzel:
@rageagainsthepc i fixed another bug in the Xen read implementation :)
we were not reading correctly when the request was _inside_ a page, and not just right a the beginning of it.
it's fixed now, and volatility works

Capture d’écran de 2021-03-18 13-49-09

Issue #167, opened 2021-02-19
KVM: Handle read size > 4K


Wenzel:
From #161 , Volatility3 is performing huge read operations, far beyong the usual 4K.
However, in KVM [`invalid_page_access`](https://github.com/KVM-VMI/kvm/blob/0c4bf8391af19a442936322a06d441d5bf1ab298/virt/kvm/introspection/kvmi_msg.c#L327) returns true if the size requested is superior to the `PAGE_SIZE`, which is 4K.

This means that the `libmicrovmi` should split the read into 4K chunks.

PR #172, opened 2021-02-22
KVM: handle any read size

Wenzel:
fix https://github.com/Wenzel/libmicrovmi/issues/167


Wenzel:
Confirm this works with mem-dump using a buffer of `1024 * 1024` bytes on KVM driver

PR #173, opened 2021-02-22
CI: Improve speed with actions/cache

Wenzel:


Wenzel:
* `debian_package` reduced from `4m40` to `2m20`
*  `release`: unknown

Issue #177, opened 2021-03-02
Include libmicrovmi.h generated C header into the Debian package


Wenzel:
fixed by https://github.com/Wenzel/libmicrovmi/pull/176

Wenzel:
The publish job fails because of this
https://github.com/Wenzel/libmicrovmi/runs/2016056690#step:4:214

"Source directory was modified by build.rs during cargo publish. Build scripts should not modify anything outside of OUT_DIR."

Wenzel:
finding assets in OUT_DIR is not supported by cargo deb, and there is an opened issue:
https://github.com/mmstick/cargo-deb/issues/154

PR #179, opened 2021-03-03
cargo: fix location of generated file libmcrovmi.h

Wenzel:
fix https://github.com/Wenzel/libmicrovmi/issues/177

There is standard solution avaialble right now, they are discussed on the rust-lang.
This should be considered a workaround in the meantime

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,14 +1,31 @@
 //! This build script will run cbindgen to generate libmicrovmi C header file
 
 use std::env;
-use std::path::Path;
+use std::fs;
+use std::path::PathBuf;
 
 use cbindgen::Config;
 
 fn main() {
     let crate_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -21,5 +38,5 @@ fn main() {
 
     println!("cargo:rerun-if-changed=src/capi.rs");
     // if it has been removed
-    println!("cargo:rerun-if-changed=c_examples/libmicrovmi.h");
+    println!("cargo:rerun-if-changed=target/capi/libmicrovmi.h");
```
rageagainsthepc, line 41: I am not familiar with the `build.rs` workflow, but is this path correct? Shouldn't it be `target/release/capi/libmicrovmi.h`?

Wenzel: oh, it's not correct indeed.
i should add 2 lines for target/debug/xxx and target/release/xxx

PR #195, opened 2021-04-23
Memory as file descriptor

Wenzel:
This PR implements the memory as a file descriptor, using Rust standard traits
- `Read`
- `Write`
- `Seek`

A second memory object is implemented to handle "padded" memory reads, useful for Volatility.

Some unit tests have been implemented for the seek implementation.

### DIFF
--- a/src/capi.rs
+++ b/src/capi.rs
@@ -4,6 +4,7 @@ use bitflags::_core::ptr::null_mut;
 use cty::{c_char, size_t, uint16_t, uint64_t, uint8_t};
 use std::convert::TryInto;
 use std::ffi::{c_void, CStr, CString};
+use std::io::{Read, Seek, SeekFrom};
 use std::slice;
 
 /// Support passing initialization options
@@ -78,15 +79,15 @@ pub unsafe extern "C" fn microvmi_destroy(context: *mut c_void) {
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe extern "C" fn microvmi_pause(context: *mut c_void) -> bool {
-    let driver = get_driver_mut_ptr(context);
-    (*driver).pause().is_ok()
+    let mut driver = get_driver_mut_ptr(context);
+    driver.pause().is_ok()
 }
 
 #[allow(clippy::missing_safety_doc)]
 #[no_mangle]
 pub unsafe extern "C" fn microvmi_resume(context: *mut c_void) -> bool {
-    let driver = get_driver_mut_ptr(context);
-    (*driver).resume().is_ok()
+    let mut driver = get_driver_mut_ptr(context);
+    driver.resume().is_ok()
 }
 
 #[allow(clippy::missing_safety_doc)]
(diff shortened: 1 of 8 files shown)


Wenzel:
This PR is ready for review.
@rageagainsthepc , if you want to have a quick look.

rageagainsthepc:
Imho the microvmi struct looks a bit overly complicated. I would prefer a design where you (as a library user) initialize the driver like we used to and then pass it to one of the memory abstraction classes:
```rust
let drv = driver_init();
let memory = Memory::new(drv);
let padded_memory = Paddedmemory::new(drv);
```
Not every user needs memory abstractions, but those who do could easily initialize them.
But it's just a though, haven't looked at the PR in detail yet. Maybe I'm not seeing the bigger picture.

Wenzel:
Hi @rageagainsthepc and thank you for your feedback.

I should have explained a bit more why I decided this implementation and design when opening the PR.
I would have also have prefered a simpler design, where the library would return a trait object (`Box<dyn Introspectable`), and provide wrapper on top of this, (`Memory::new(drv)`).

The issues comes from exposing this API in C and Python. because how do you keep the context ?

The C API exposes a `void* drv` pointer, to be casted into our `box<dyn Introspectable>`, however, this wouldn't retain the `Memory` object context as well, since it's a separate object.

That's the main reason why I choose to go this way, keep a single struct `Microvmi` to be casted as `void*` into the C world, back and forth, and able to offer the read/write/seek abstraction underneath.

A second reason is that we are able to simplify our read implementations in the drivers, mainly the Xen driver, since the main algorithm to iterate over each frame is implemented in the layer above.

And another reason is to be able to implement batch read operation later on, hidden in the Memory object implementation.
Memflow has support for this since some of its targets have a very slow latency, so have to batch you read operations, otherwise the whole read will become slow and uneffective.

src/capi.rs
```diff
@@ -97,19 +101,25 @@ pub unsafe extern "C" fn microvmi_read_physical(
...
 
-    let mut bytes_read_local = 0;
-    let res = (*driver)
-        .read_physical(
-            physical_address,
-            slice::from_raw_parts_mut(buffer, size),
-            &mut bytes_read_local,
-        )
-        .is_ok();
+    if driver
```
rageagainsthepc, line 106: This file was supposed to be a simple C abstraction layer. Augmenting it with code that goes beyond that is bad design imho.

src/memory.rs
```diff
@@ -0,0 +1,235 @@
...
+
+    // Seek
+    #[test]
+    fn test_seek_start() -> Result<()> {
+        let max_addr: u64 = 1000;
+        let mut mock_introspectable = MockIntrospectable::new();
+        mock_introspectable
+            .expect_get_max_physical_addr()
+            .returning(move || Ok(max_addr));
+        let mut memory = Memory::new(Rc::new(RefCell::new(Box::new(mock_introspectable)))).unwrap();
```
rageagainsthepc, line 190: I would welcome it if we followed the arrange, act, assert pattern as it makes the tests more readable imo. https://automationpanda.com/2020/07/07/arrange-act-assert-a-pattern-for-writing-good-tests/
I don't think that you should add comments like the article does, but simply adding newlines between the tree blocks would be helpful.

src/memory.rs
```diff
@@ -0,0 +1,235 @@
...
+        // seek 0 doesn't move position
+        assert_eq!(0, memory.seek(SeekFrom::Start(0))?);
```
rageagainsthepc, line 192: Don't put multiple unrelated asserts into the same test case, if you want to avoid code duplication use the `test-case` crate. See https://github.com/Wenzel/libmicrovmi/blob/master/src/driver/kvm.rs#L398

src/microvmi.rs
```diff
@@ -0,0 +1,207 @@
...
+    pub padded_memory: PaddedMemory,
+}
+
+impl Microvmi {
+    /// Initializes a new Microvmi instance
+    ///
+    /// # Arguments
+    ///
+    /// * `domain_name` - The domain name
+    /// * `driver_type` - The driver type to initialize. None will attempt to initialize every driver avaiable
```
rageagainsthepc, line 40:
```suggestion
    /// * `driver_type` - The driver type to initialize. None will attempt to initialize every driver available
```

src/microvmi.rs
```diff
@@ -0,0 +1,207 @@
...
+        let ref_drv = Rc::new(RefCell::new(drv));
+        Ok(Microvmi {
+            drv: ref_drv.clone(),
+            memory: Memory::new(ref_drv.clone())?,
+            padded_memory: PaddedMemory::new(ref_drv.clone())?,
+        })
+    }
+
+    /// Retrieve the number of VCPUs.
+    pub fn get_vcpu_count(&self) -> Result<u16, Box<dyn Error>> {
```
rageagainsthepc, line 88: You basically implement the `Introspectable` trait without officially implementing the `Introspectable` trait. That's nasty. :/

rageagainsthepc: It's also a lot of code duplication and unnecessary indirection. What's so wrong about making the `drv` public in order to allow direct access?

rageagainsthepc (changes requested):
I would at least improve the tests. As for the other comments: I understand if you are set on your current approach, I'll leave that up to you.

PR #198, opened 2021-05-21
Add Integration Tests for KVM

Wenzel:
This PR adds integration tests for the KVM driver


tests/README.md
```diff
@@ -0,0 +1,25 @@
+# Integration Tests
+
+Instructions for all tests:
+- Update the values in `tests/config.rs` according to your environment
+- `cargo test -- --test-threads=1`: the tests must be run sequentially since they share the same VM
```
rageagainsthepc, line 5: https://lib.rs/crates/serial_test_derive

tests/kvm.rs
```diff
@@ -0,0 +1,201 @@
...
+    INIT.call_once(|| {
+        env_logger::builder().is_test(true).init();
+    });
+    // setup before test
+    setup_test();
+
+    // setup test execution in a thread
+    let (done_tx, done_rx) = mpsc::channel();
+    let handle = thread::spawn(move || {
+        let val = test();
```
rageagainsthepc, line 32: What happens if a test assert fails? Won't this simply panic and prevent you from executing your teardown? Wouldn't you need something like `panic::catch_unwind()` in order to prevent that?

rageagainsthepc: https://medium.com/@ericdreichert/test-setup-and-teardown-in-rust-without-a-framework-ba32d97aa5ab

tests/kvm.rs
```diff
@@ -0,0 +1,201 @@
...
+use std::sync::{mpsc, Once};
+use std::thread;
+use std::time::Duration;
+
+use config::{KVMI_SOCKET, TIMEOUT, VIRSH_URI, VM_NAME};
+
+// to init env logger
+static INIT: Once = Once::new();
+
+fn run_test<T>(test: T) -> ()
```
rageagainsthepc, line 17: What about: https://docs.rs/test-context/0.1.2/test_context/

tests/kvm.rs
```diff
@@ -0,0 +1,201 @@
...
+    // setup test execution in a thread
+    let (done_tx, done_rx) = mpsc::channel();
+    let handle = thread::spawn(move || {
+        let val = test();
+        done_tx.send(()).expect("Unable to send completion signal");
+        val
+    });
+
+    // wait for test to complete until timeout
+    let timeout = Duration::from_secs(TIMEOUT);
```
rageagainsthepc, line 38: What about: https://docs.rs/ntest/0.3.3/ntest/attr.timeout.html

rageagainsthepc: There is also a possibility to set the timeout globally if you are simply worried about a blocking CI run. https://os.phil-opp.com/testing/#timeouts

rageagainsthepc:
Very cool btw 👍

tests/kvm.rs
Wenzel: Thanks for the suggestion.
The issue i had with ntest is that the timeout is applied to the whole test, including setup and teardown.


However, if i can combine that with test-context, it can work

rageagainsthepc: Why is it a problem that setup and teardown are included in the timeout? Just account for that by increasing your timeout accordingly. 🤷‍♂️

Wenzel: The setup time depends on the VM that's being tested.
It's not the same thing to revert a winxp 128MB snapshot and an 8GB RAM Windows 10 snapshot :)
And I didn't wanted to make the timeout depend on the environment, if possible

tests/kvm.rs
Wenzel: Interesting crate, but it doesn't seem well known in the community ?
the repo has 4 :star: and the last commit was 4 months ago, so I'm not convinced yet  :)

tests/kvm.rs
Wenzel: `thread_spawn` will catch anything panicking in the thread's execution.
The result can be obtained when you call join():
https://doc.rust-lang.org/std/thread/struct.JoinHandle.html#method.join

and the teardown code is still executed
example with a failing test:
~~~rust
    fn test_init_driver() {
        run_test(|| {
            assert!(false);
            init_driver();
        })
    }
~~~

`RUST_LOG=debug cargo test --features kvm test_init_driver -- --nocapture`
output
~~~
running 1 test
[2021-05-25T06:27:02Z DEBUG kvm] setup test
thread '<unnamed>' panicked at 'assertion failed: false', tests/kvm.rs:102:13
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
[2021-05-25T06:27:03Z DEBUG kvm] teardown test
thread 'tests::test_init_driver' panicked at 'Test timeout: Disconnected', tests/kvm.rs:43:9
test tests::test_init_driver ... FAILED
~~~

tests/kvm.rs
rageagainsthepc: Yeah, that makes sense.

tests/kvm.rs
rageagainsthepc: Hm, still better than implementing everything yourself, I think. ;) You are right, it's not well known, but it looks decent to me at first glance (although the macros were a hard pass for me^^). It also has very little functionality which might explain why it is not actively developed. Moreover, it's not like we would be using it in production code.
I personally like how this crate let's you structure your test harness because it's more in line with what I am used to from well established test frameworks in other languages than the custom harness. Plus, it would be neat if you could use this together with ntest so you don't have to implement context cancellation yourself.

rageagainsthepc: To be clear, I totally understand if that's not convincing enough for you ;)

Wenzel:
Updated with a custom harness implementation based on:
https://www.fluvio.io/blog/2021/04/rust-custom-test-harness/

Wenzel:
Merging this

PR #200, opened 2021-05-22
Make sure latest stable rust toolchain is installed and made default

rageagainsthepc:


.gitpod.yml
```diff
@@ -6,9 +6,7 @@ tasks:
 
 vscode:
   extensions:
-    # Theia does not support rust-analyzer, therefore this plugin list targets vscode instead.
```
rageagainsthepc, line 9: VSCode is the default editor now so this comment is not necessary anymore.

PR #210, opened 2021-09-03
kvm: fix bytes_read

Wenzel:
KVM driver read_physical: bytes_read value was not updated correctly


rageagainsthepc:
This would have been a good opportunity to add some unit tests which make sure that `bytes_read` is updated correctly. 😛

Wenzel:
I agree, with first integration tests are what yields most value here

PR #254, opened 2025-03-12
Fix FFI handling of optional memflow connector_args

ajkhoury:
As stated in the libmicrovmi documentation, `memflow_connector_args` are optional, however omitting the `memflow_connector_args` from the FFI causes an assertion failure in `std::slice::from_raw_parts`. The documentation for `std::slice::from_raw_parts` states that the pointer must be non-null and properly aligned, since it promises to return a slice that actually refers to valid memory. Prior to these changes, the code would attempt to pass a null pointer via `args_arr` to `std::slice::from_raw_parts` if the optional `memflow_connector_args` were not provided.

### DIFF
--- a/src/capi/params.rs
+++ b/src/capi/params.rs
@@ -88,20 +88,28 @@ impl TryFrom<DriverInitParamsFFI> for DriverInitParams {
                 MemflowConnectorParamsFFI::Default {
                     args_arr,
                     args_arr_len,
-                } => unsafe {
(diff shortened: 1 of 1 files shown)


ajkhoury:
Thanks for the quick response @Wenzel. Sorry I forgot to rustfmt the changes! I rebased from latest master and formatted the changes, let me know if there's anything else that needs changed :rocket:

